PR #5, opened 2016-07-29
Ng2 bootstrap

frankfu03:
Please review and merge it to master if it works fine. Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,8 @@
     "rxjs": "5.0.0-beta.6",
     "systemjs": "0.19.27",
     "zone.js": "^0.6.12",
-    "bootstrap": "3.3.7"
+    "bootstrap": "3.3.7",
+    "ng2-bootstrap": "^1.0.24",
+    "moment": "^2.14.1"
   }
 }
--- a/tools/config/project.config.ts
+++ b/tools/config/project.config.ts
@@ -22,6 +22,7 @@ export class ProjectConfig extends SeedConfig {
       ...this.NPM_DEPENDENCIES,
       // {src: 'jquery/dist/jquery.min.js', inject: 'libs'},
       // {src: 'lodash/lodash.min.js', inject: 'libs'},
+      {src: 'ng2-bootstrap/bundles/ng2-bootstrap.js', inject: 'libs'},
       {src: 'bootstrap/dist/css/bootstrap.min.css', inject: true}, // inject into css section
     ];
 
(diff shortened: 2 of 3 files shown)


oehm-smith:
I don't know it that well, but it looks odd that you need to include it twice.  Are you sure both are required.  I would have thought the one in the project config would have been all that is required.  Could you confirm?

frankfu03:
Yeah, it is odd. I was asked to do a merge before doing rebase. So there are duplicate comments. There are only 4 lines added to 3 files.

frankfu03:
"ng2-bootstrap" and "moment" were added to package.json for installing dependencies into node_modules; "ng2-bootstrap/ng2-bootstrap" added to tools/config/seed.config.ts
is used for importing modules from ng2-bootstrap in JS. So I can confirm that all 4 lines added are needed.

oehm-smith:
So with things how they are (after rebase / merge), can you confirm you do need to modify the seed.config.ts - as I understood it, to customise you only modify the project.config.ts.  That is what the angular2-seed doco says.  In https://github.com/mgechev/angular2-seed/wiki/Architecture-and-usage-of-angular2-seed is 

### Config

The project defines a base configuration inside: tools/config/seed.config.ts. The config can be easily overridden in: tools/config/project.config.ts.

So could you double-check you need that modification.  And if you do can you make it to the project config?

frankfu03:
Ok. I may remove the new line "'ng2-bootstrap/ng2-bootstrap':  ..." from seed.config.ts. It seems the UI still works fine after having removed it. I'll double check it

But we should be aware that "'ng2-bootstrap/ng2-bootstrap': 'node_modules/ng2-bootstrap/bundles/ng2-bootstrap.js'," in seed.config.ts is defining a path that is to be used for "import {DROPDOWN_DIRECTIVES} from 'ng2-bootstrap/ng2-bootstrap';" This idea comes from the Angular2 seed project which defined "'@angular/forms': `node_modules/@angular/forms/bundles/forms.umd.js`," for "import { REACTIVE_FORM_DIRECTIVES } from '@angular/forms';" in home.component.ts.

oehm-smith:
If needed, can it go in the project.config.ts instead?

frankfu03:
It seems that it cannot be defined in project.config.ts without significant changes. Because ProjectConfig extends from SeedConfig, we may merge both files if it is necessary. 

The project.config.ts defines .css and .js files to be injected into index.html, while seed.config.ts defines global variables and SYSTEM_CONFIG variables. 

I had another check and found out that the new line "'ng2-bootstrap/ng2-bootstrap': ..." added to seed.config.ts is actually injected into index.html as part of the System.config.

We may discuss this issue on Monday.

PR #12, opened 2016-08-03
Typescript style improvements

oehm-smith:


lbodor:
Rebase

src/client/service-worker-controller.ts
```diff
@@ -2,11 +2,11 @@
 // - found Events to use by looking at typings/globals/service_worker_api/index.d.ts and
 //   looking for the 'on' event functions.
 
-var debugEvent = function (event:any) {
+function debugEvent (event:any): void {
```
lbodor: What is the type of event?

src/client/service-worker-controller.ts
```diff
@@ -45,16 +45,16 @@ var cleanCache1 = function (event:any) {
   );
 };
 
-self.addEventListener('install', function (event:InstallEvent) {
+self.addEventListener('install', (event:InstallEvent) => {
   debugEvent(event);
 });
```
lbodor: Could these be written as self.addEventListener('install', debugEvent)?

oehm-smith:
It appears to work.  Code updated.

PR #14, opened 2016-08-04
Add 'Select Site' tab for searching and selecting CORS sites

frankfu03:


frankfu03:
For testing purpose, allow users to retrieve all sites from DB when both Site Name and Four Char ID input fields are empty; shows the four char ID and name of the selected site on the same page/tab, instead of going to next page/tab as it should do.

frankfu03:
I have updated the branch based on the review feedback. Please let me know if I missed any items. In addition, I added a ajax loading image for searching.

PR #15, opened 2016-08-12
Implement Service Worker and Web app caching

oehm-smith:
* Service Worker caching integrated with select-site UI code and provides
button to enable new version of site to be installed on client


src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,82 @@
+import {Injectable, NgZone} from '@angular/core';
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+
+/**
+ * This class provides the service for the application's service worker that is global to the browser.
+ */
+@Injectable()
+export class ServiceWorkerService {
+  // Setup Observables (BehaviorSubject's to be specific) to communicate between components
+  private _clearCacheFlag = new BehaviorSubject<boolean>(false);
```
oehm-smith: _ not advised

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        if (event.data.error) {
+          if (failureCallback) {
+            failureCallback();
+          } else {
+            throw new Error('ERROR clearing cache');
+          }
+        } else {
+          if (successCallback) {
+            // Advise any Observers that the cache has been changed so they can retrieve the current list via getCacheList()
+            _this._clearCacheFlag.next(true);
```
oehm-smith: Move outside guard

oehm-smith:
Code updated and ready for new review.  Add comments against lines in source code as required.

src/client/app/+home/home.component.ts
```diff
@@ -1,7 +1,9 @@
-import { Component, OnInit } from '@angular/core';
-import { REACTIVE_FORM_DIRECTIVES } from '@angular/forms';
+import {Component, OnInit} from '@angular/core';
```
lbodor: Use spaces inside these curly braces, as in { Component, OnInit }

src/client/app/+home/home.component.ts
```diff
@@ -56,4 +71,26 @@ export class HomeComponent implements OnInit {
     return false;
   }
 
+  /**
+   * Component method to request the Service Worker clears it's cache.
+   */
+  clearCache=() => {
```
lbodor: function clearCache(): void {

src/client/app/+home/home.component.ts
```diff
@@ -56,4 +71,26 @@ export class HomeComponent implements OnInit {
...
+    }, (error: Error) => {
+      throw new Error('Error in clearCache: '+ error.message);
+    });
+  };
+
+  updateCacheList=() => {
+    this.serviceWorkerService.getCacheList().then((data: string[]) => {
+      this.cacheItems.length = 0;
+      this.cacheItems = data;
+      this.numberCacheItems = this.cacheItems.length;
```
lbodor: Why is numberCacheItems needed? Perhaps it would be better to use cacheItems.length instead.

src/client/app/+home/home.component.ts
```diff
@@ -56,4 +71,26 @@ export class HomeComponent implements OnInit {
...
+    }, (error: Error) => {
+      console.warn('Error in updateCacheList: '+ error.message);
+    }).catch((error: any) => {
```
lbodor: Why any?

src/client/app/+home/home.component.ts
```diff
@@ -56,4 +71,26 @@ export class HomeComponent implements OnInit {
...
+      console.error("Caught error in updateCacheList:", error);
+    });
+  };
```
lbodor: Why is one error logged as a warning and other as an error? Isn't it unnecessary to supply both a reject handler and a catch handler?

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * This class provides the service for the application's service worker that is global to the browser.
+ */
+@Injectable()
+export class ServiceWorkerService {
+  // Setup Observables (BehaviorSubject's to be specific) to communicate between components
+  private clearCacheFlag = new BehaviorSubject<boolean>(false);
+  // Observable
+  public clearCache$ = this.clearCacheFlag.asObservable();
```
lbodor: clearCache

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Generic method to post messages to the service worker and return a promise to the caller
+   * @param operation to be performed
+   * @param message is the optional message to be sent with that operation
+   * @returns {Promise<T>} that is appropriate to the client of this method
+   */
+  postMessage(operation: string, message?: string): Promise<any> {
+    let messageObject = {operation: operation, message: message} as MessageObject;
+    return new Promise(function (resolve: any, reject: any) {
```
lbodor: Why any?

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this.postMessage('get_cache');
+  }
+
+  /**
+   * Generic method to post messages to the service worker and return a promise to the caller
+   * @param operation to be performed
+   * @param message is the optional message to be sent with that operation
+   * @returns {Promise<T>} that is appropriate to the client of this method
+   */
+  postMessage(operation: string, message?: string): Promise<any> {
```
lbodor: Since you have a message interface, you should use it in the method signature as well.

src/client/app/shared/toolbar/toolbar.component.ts
```diff
@@ -9,4 +11,50 @@ import { Component } from '@angular/core';
...
+
+  updateCacheList=() => {
+    this.serviceWorkerService.getCacheList().then((data: string[]) => {
+      this.cacheItems.length = 0;
+      this.cacheItems = data;
+      this.numberCacheItems = this.cacheItems.length;
+    }, (error: Error) => {
+      console.warn('Error in updateCacheList: '+ error.message);
+    }).catch((error: any) => {
+      console.error("Caught error in updateCacheList:", error);
```
lbodor: As above, there is probably no need for a reject handler as well as a catch handler.

src/client/service-worker-controller.ts
```diff
@@ -2,74 +2,102 @@
...
+  event.waitUntil(
+    self.caches.open(cacheName)
+      .then((cache: Cache) => {
+        cache.add('/');
+      })
+  );
+});
 
-self.addEventListener('activate', (event:ExtendableEvent) => {
+self.addEventListener('activate', function (event: ExtendableEvent) {
```
lbodor: Use fat arrow notation for lambdas.

src/client/service-worker-controller.ts
```diff
@@ -84,7 +112,59 @@ self.addEventListener('notificationclick', debugEvent);
...
 self.addEventListener('notificationclose', debugEvent);
 
-self.addEventListener('message', debugEvent);
+/**
+ * Return list of URLS for requests in cache
+ * @param event
+ */
+function getCache(event: MessageEvent): Promise<string[]> {
+  debugMsg('getCache: ' + cacheName);
+  let cacheUrls: any = [];
```
lbodor: Why any? Aren't they strings?

tslint.json
```diff
@@ -7,7 +7,7 @@
     "indent": ["spaces"],
     "max-line-length": [true, 140],
     "member-ordering": [true,
-      "public-before-private",
+      "XXXpublic-before-private",
```
lbodor: What is this for?

src/client/service-worker-controller.ts
```diff
@@ -84,7 +112,59 @@ self.addEventListener('notificationclick', debugEvent);
...
+      reject(error);
+    });
+  });
+}
+
+self.addEventListener('message', (event: MessageEvent) => {
+  debugEvent(event);
+  let messageObject: MessageObject = event.data;
+  let op: string = messageObject.operation;
+  let message: string = messageObject.message;
```
lbodor: Do you know about deconstruction syntax?
let {operation, message} = event.data.messageObject;

src/client/service-worker-controller.ts
```diff
@@ -2,74 +2,102 @@
...
+  debugMsg('  deleteCache function - clean cache: ' + cacheToDelete);
+  event.waitUntil(
+    self.caches.keys().then((cacheNames: string[]) => {
+      // WARNING - Intellij says there is a syntax error here but AFIACT there isn't.  Don't spent time looking at this.
+      return Promise.all(cacheNames.filter((cacheToTest: string) => {
+          debugMsg('    cache: ' + cacheToTest);
+          return cacheToTest === cacheToDelete;
+        }).map((cacheToDelete: string) => {
+          self.caches.delete(cacheToDelete);
+        })
```
lbodor: eta-reduction let's you write this map as
```map(self.caches.delete)```

File: src/client/app/+home/home.component.ts (hunk 6 above)
Comment (oehm-smith) [inline src/client/app/+home/home.component.ts]: https://github.com/Microsoft/TypeScript/issues/6283 "Promise catch error parameter should not be of type any" is still open.  Any is the only possible way at the moment.

File: src/client/service-worker-controller.ts [hunk 17]
```diff
@@ -2,74 +2,102 @@
...
   );
-};
+}
+
+/**
+ * Delete the named cache
+ * @param event
+ * @param cacheToDelete to delete
+ */
+function deleteCache(event: any, cacheToDelete: string): void {
```
lbodor: Since `deleteCache` does not really depend on the event. Since it the caller of deleteCache is responding to an event, perhaps it should call `event.waitUntil(deleteCache(cacheToDelete))`.

src/client/app/+home/home.component.ts
oehm-smith: Nice spot re: warn should be error (it used to be like this for a reason). 

Promises can be rejected and also anywhere on the call chain an exception can be thrown and they aren't picked up by the reject().

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,66 @@
...
+  // Setup Observables (BehaviorSubject's to be specific) to communicate between components
+  private clearCacheFlag = new BehaviorSubject<boolean>(false);
+  // Observable
+  public clearCache$ = this.clearCacheFlag.asObservable();
+
+  /**
+   *
+   * @returns {Promise<any>} that may contain a message about the operation
+   */
+  clearCacheService(): Promise<string> {
```
lbodor: Why `clearCacheService` as method name? Why not just `clearCache`?

src/client/app/shared/service-worker/service-worker.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    let promise: Promise<string> = this.postMessage('clear_cache');
+
+    promise.then(() => {
+      this.clearCacheFlag.next(true);
+    });
+    return promise;
```
lbodor: Is this a bug? Does `then` modify the promise or return a new promise?

src/client/app/+home/home.component.ts
oehm-smith: Only having a catch() seems to work ok.

src/client/app/shared/service-worker/service-worker.service.ts
oehm-smith: Changed to `Function`, which seems to be the correct type.

tslint.json
oehm-smith: I commented out to see if will stop an error from happening during tslint that prevented the build.  Does and will remove this.

src/client/service-worker-controller.ts
oehm-smith: No it didn't like it.  If it did work I don't like it as the terseness takes away from the readability.

src/client/service-worker-controller.ts
oehm-smith: Then that would mean that everywhere I needed to call deleteCache I had to wrap it in this.  I dont' believe it buys us anything.

src/client/app/shared/service-worker/service-worker.service.ts
oehm-smith: Historical when i was developing and I've been meaning to change!

src/client/app/shared/service-worker/service-worker.service.ts
oehm-smith: Its so the `clearCacheFlag.next()` can run in `clearCache()`.  It appears to operate as expected with the same promise being returned.

oehm-smith:
Check it!

PR #18, opened 2016-08-18
Site info

frankfu03:


src/client/app/+site-info/site-info.component.ts
```diff
@@ -0,0 +1,234 @@
...
+   * Clear all variables/arrays
+   */
+  public ngOnDestroy() {
+    this.site = null;
+    this.siteLocation = null;
+    this.siteOwner = null;
+    this.metadataCustodian = null;
+    this.status = null;
+    this.receivers = [];
+    this.antennas = [];
```
oehm-smith: use `array.length = 0` instead.

src/client/app/+site-info/site-info.component.ts
```diff
@@ -0,0 +1,234 @@
...
+      if (isOpen) {
+        return null;
+      }
+    }
+    return true;
+  }
+
+  public getAllCorsSetupsBySiteId(siteId: number) {
+    this.corsSetupService.getCorsSetupsBySiteId(siteId).subscribe(
+      (responseJson: any) => {
```
oehm-smith: Let's pass around objects so we can make the response a type.  So below can be:

    for (let setupObj : Setup of responseJson._embedded.setups) { ...

etc...

src/client/app/+site-info/site-info.component.ts
```diff
@@ -0,0 +1,234 @@
...
+        this.receivers.sort(this.compareDate);
+        this.antennas.sort(this.compareDate);
+
+        // Current receiver/antenna (even null) are the first item in the arrays and open by default
+        this.receivers.unshift(currentReceiver);
+        this.antennas.unshift(currentAntenna);
+        this.status.isReceiversOpen.unshift(true);
+        this.status.isAntennasOpen.unshift(true);
+      },
+      (error: any) =>  this.errorMessage = <any>error
```
oehm-smith: Type type can be '`Error`'.

    (error: Error) => ...

src/client/app/+site-info/site-info.component.ts
lbodor: Please put your comment below the code, not above it.

src/client/app/+site-info/site-info.component.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+    return true;
+  }
+
+  public getAllCorsSetupsBySiteId(siteId: number) {
+    this.corsSetupService.getCorsSetupsBySiteId(siteId).subscribe(
+      (responseJson: any) => {
+        let currentReceiver: any = null;
+        let currentAntenna: any = null;
+        for (let setupObj of responseJson._embedded.setups) {
```
lbodor: Prefer setup to setupObj. Most things are objects, no need for suffix.

src/client/app/shared/cors-site/cors-site.service.ts
```diff
@@ -22,23 +23,25 @@ export class CorsSiteService {
 
   /**
    * Returns an Observable for the HTTP GET request for the REST Web Service resource.
+   * @param {string} fourCharacterId - The Four Character Id of the site.
+   * @param {string} siteName - The name of the site.
    * @return {object[]} The Observable for the HTTP request.
    */
   getCorsSitesBy(fourCharacterId: string, siteName: string): Observable<any> {
```
oehm-smith: lets type the returns

    Observable<CorsSite[]>

oehm-smith:
Page looks great.  Here is a summary of feedback:

1. Lets introduce types.  I'm not sure if you can return a type from an `http.get` since I think that is a general json object with fields like 'embedded' and 'page'.  But you subscribe to these generally and assign to something after extracting from the `json._embedded` so that is definitely do-able.

2. Testing - we need to do more of this.  The hardest part is starting!

3. UI Feedback not intended for here but this is a place to record it to consider in future sprints:

* Return key should default to 'search'
* We should have responsive feedback where as the user types a list of possible choices is presented

src/client/app/+site-info/site-info.component.ts
lbodor: Let's worry about types after we switch to geodesyml and jsonix.

PR #26, opened 2016-08-26
Update to angular seed latest ng rc5

oehm-smith:
Pushing so all can have an opportunity to see the state of upgrading to latest Angular2-seed with NG rc.5

Works in local dev but not on prod (aws s3).  Some NgModules problems.


lbodor:
Let's get his branch working first.

PR #28, opened 2016-08-31
Add auto-height attribute directive to ensure header always shown

frankfu03:


src/client/app/shared/global/auto-height.directive.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  constructor(private elem: ElementRef) {}
+
+  ngOnInit() {
+    let bodyHeight = window.innerHeight;
+    this.resize(bodyHeight);
+  }
+
+  @HostListener('change', ['$event'])
+  onChange(event: any) {
```
lbodor: Is it possible to specify a specific event type here?

PR #34, opened 2016-09-13
Create jsonix service

oehm-smith:
Create jsonx service and wire up

    * SiteInfoService method getSiteLogByFourCharacterId() duplicated
    to getSiteLogByFourCharacterIdGeodesyMl() that uses the JSONIX
    service where the data is GeodesyMl.  This is not called
    and we can change to use this when we are ready.
    * todo - marshaller back to XML


src/client/app/shared/jsonix/jsonix.service.ts
```diff
@@ -0,0 +1,49 @@
...
+  // WS_URL : string = 'https://dev.geodesy.ga.gov.au'; // dev
+  static WS_URL: string = 'https://dev.geodesy.ga.gov.au'; // test
+
+  /**
+   * Given valid GeodesyML instatnce, use translate to JSON.
+   *
+   * @param geodesyMl valid input to translate
+   * @return The  JSON representation of the given geodesyMl
+   */
+  geodesyMlToJson(geodesyMl: string): string {
```
lbodor: We spell two-letter abbreviations in capitals, so it's geodesyMLToJson.

src/client/app/shared/site-log/site-log.service.ts
```diff
@@ -3,76 +3,117 @@ import { Http, Response } from '@angular/http';
...
  */
 @Injectable()
 export class SiteLogService {
-  // WS_URL : string = 'http://localhost:8080/geodesy-web-services';
-  // WS_URL : string = 'https://dev.geodesy.ga.gov.au'; // dev
-  WS_URL : string = 'https://dev.geodesy.ga.gov.au'; // test
+  // WS_URL: string = 'http://localhost:8080/geodesy-web-services';
+  WS_URL : string = 'https://dev.geodesy.ga.gov.au'; // dev
+  // WS_URL: string = 'https://dev.geodesy.ga.gov.au'; // test
+  that: any = this;
```
lbodor: Is "that" used?

PR #40, opened 2016-09-14
Switch to GeodesyML/JSON

frankfu03:


src/client/app/app.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component } from '@angular/core';
 // import { HTTP_PROVIDERS } from '@angular/http';
 import { Config, CorsSiteService, CorsSetupService, SiteLogService, GlobalService, NameListService,
-          ServiceWorkerService, JsonixService } from './shared/index';
+          AutoHeightDirective, ServiceWorkerService, JsonixService } from './shared/index';
```
oehm-smith: I fixed the AutoHeight in https://github.com/GeoscienceAustralia/gnss-site-manager/pull/37/files.  I don't think this way will work.

oehm-smith:
How about using a different branch name that describes the work done.

PR #54, opened 2016-09-27
Global save

frankfu03:


oehm-smith:
How about some tests?  Such as confirm that upon save() that the backend is hit (use a spy).

oehm-smith:
The button down the bottom has (un) available emphasis but the one on the toolbar doesn't.

PR #56, opened 2016-09-30
Wfs query service

oehm-smith:
* Defined WFS Service and wired up to allow wildcards in searches
* Only 'four character id' implemented
* Some debug kept in code


oehm-smith:
Fixed tests; right to be reviewed and merged now.

PR #60, opened 2016-10-11
Split gnss receiver component

Layoric:
This project is looking great, good work :+1: .

One thing I've noticed is that the `site-info` component is getting a bit large and probably should act as more of an orchestration component splitting up the different panels into separate components.

This PR splits out the GNSS receiver info panel into it's own component which `site-info` uses. Eg

``` html
<gnss-receiever-info [status]="status" [gnssReceivers]="siteLogModel.gnssReceivers"></gnss-receiever-info>
```

Any feedback would be appreciated :smile:


frankfu03:
Thanks Darren for your great support. It is a good idea to split out receiver info from site info as it is growing larger. We have been thinking of doing it to both receiver and antenna info, but no action so far. Many thanks again.

Layoric:
@frankfu03 No worries. I'll still be going through this project tomorrow, are you happy for me to split out some of the other panels tomorrow (in separate pull requests of course) or should I leave those for yourself or others to do?

frankfu03:
We would be more than happy to have you to help us more. But meanwhile, we do not want to make yourself too busy with our project as it is your last a few days at GA. So please do whatever you can when you are available to help us.

PR #66, opened 2016-10-12
Lint

lbodor:


Layoric:
Just as a note, `===` for null misses `undefined` which is common. Might want to consider changing the lint rules to allow/ignore `== null`, most lint libraries can make this distinction. IMO writing `foo !== null && foo !== undefined` is overkill and letting coercion do it's thing in this case is safer and more concise.

frankfu03:
Yeah, I had a case which needs '==' for comparison (eg, 1 == '1'), but "npm run build.prod" raises it as an error.

lbodor:
OK, I'll try to re-configure the linter not to complain when comparing against null and undefined.

Layoric:
@frankfu03 I think the tslint is right to raise `1 == '1'` as an issue as there is a type change involved which for more complex comparisons can be a really annoying issue. `foo == null` is a special case due to JS/TS having a distinction between a object property never being assigned/created (undefined) and one that exists but doesn't have a value (null).

WebStorm and other IDEs can take some of these *Lint/Hint configs and incorporate them into the static analysis as well which might make for a better workflow for those using it.

PR #68, opened 2016-10-12
Save diff

frankfu03:


package.json
```diff
@@ -122,12 +122,12 @@
...
-    "@angular/core": "2.0.0",
-    "@angular/forms": "2.0.0",
+    "@angular/common": "2.0.2",
+    "@angular/compiler": "2.0.2",
+    "@angular/core": "2.0.2",
+    "@angular/forms": "2.0.2",
     "@angular/http": "2.0.0",
-    "@angular/platform-browser": "2.0.0",
+    "@angular/platform-browser": "2.0.2",
     "@angular/platform-browser-dynamic": "2.0.0",
```
Layoric: Careful letting `@angular` dependencies out of sync with one another. Ideally for the main libraries (outside of `@angular/router` etc) they should keep in lockstep with one another. IE, if you upgrade `@angular/core` from `2.0.0` to `2.0.2`, you should update the other main libs to `2.0.2` as well.

frankfu03:
Yes, you are right. I will do it now. 

What I have done is based on "npm install" command which suggests me to upgrade these @angular components to specific version in matching with ng2-bootstrap version. Those not upgraded were not mentioned.

PR #78, opened 2016-10-21
Refactor out global service

oehm-smith:
* create HttpUtilsService
* create ConstantsService
* use Routes not selectedSiteId
* remove unused fields
* renamed global.service to misc-utils.service

All tests pass.


frankfu03:
Looks all good. Well done.

We may need to add "export * from './http-utils.service';" to src/client/app/shared/global/index.ts.

PR #91, opened 2016-11-01
Sw cache saves

oehm-smith:
* Save saved data to the cache so immediate searches for same site returns changed data
* theoretically would work in offline mode except for 2 problems / TODOs:
  1. Current impl to save to cache done after successful save to backend service (easy fix)
  2. The geoserver wfs queries are POSTs and are never cached (nor is service worker currently able to cache POSTs).  Not so easy fix that will mean intercepting such POSTs and determining if ON or OFF Network and if OFF then pull from cache or LocalDB.  When ON save POSTs to cache or localDb.


lbodor:
Let's leave this on hold for now. Offline mode is low priority.

PR #92, opened 2016-11-04
Post review

frankfu03:


src/client/app/shared/global/misc-utils.service.ts
```diff
@@ -2,17 +2,55 @@ import { Injectable } from '@angular/core';
...
+   * Clone a JSON object from existing one so that both have no reference
+   */
+  public cloneJsonObj(obj: any): any {
+    return JSON.parse(JSON.stringify(obj));
+  }
+
+  /**
+   * Scroll the element clicked into full-view on the page and return a flag for switching block open/hide option.
+   */
+  public scrollIntoView(event: any, isBlockOpen: boolean): boolean {
```
oehm-smith: How about putting these in a ui-utils.service?

src/client/css/main.css
```diff
@@ -69,73 +75,77 @@ ul {
...
-.panel-body {
-  padding: 5px !important;
+.panel-group .panel-level-2,
+.panel-group .panel-level-2 > .panel-heading,
+.panel-group .panel-level-2 > .panel-body {
+  border-color: #bce8f1;
 }
-.panel-title {
+.panel-group .panel-title {
   font-family: "Roboto", sans-serif !important;
```
oehm-smith: This sounds like a uncommon font.  You should add a fallback to a similar but common (san-)serif.  Sorry though I can't remember how.  (Sorry that this isn't a change you made now - I just happened to see it).

src/client/app/datetime-picker/datetime-picker.component.html
```diff
@@ -1,15 +1,15 @@
 <div>
   <small [hidden]="!invalidDatetime" class="alert alert-danger">
     "{{name}}" is {{ (required && datetime.length == 0) ? "required" : "not valid" }}.
-    <br>&nbsp;&nbsp;&nbsp;&nbsp;<i>(Valid Date Format: YYYY-MM-DDThh:mm:ss.sssZ)</i>
+    <br><i>(Valid Date Format: YYYY-MM-DD hh:mm:ss)</i>
```
oehm-smith: I thought the decision was to drop seconds.  Could you confirm with Lazar / Nick?

frankfu03: We may confirm it with Nick on Monday morning meeting.

src/client/css/main.css
oehm-smith: I don't mean replace, just have a fallback that is guaranteed to exist on all devices.

PR #99, opened 2016-11-14
Disable service worker

lbodor:


src/client/index.html
```diff
@@ -49,6 +49,7 @@
   </script>
   <% } %>
 
-  <script src="js/setup-service-worker.js"></script>
```
oehm-smith: You should also remove / change the 'update application' menu item.

PR #110, opened 2016-11-17
Episodic effect

dotzhou:


oehm-smith:
@dotzhou how about squashing all these commits?  I will show you how.  Otherwise looks good.

PR #112, opened 2016-11-17
Get rid of warnings

dotzhou:


src/client/app/site-info/site-info.component.ts
```diff
@@ -625,7 +623,6 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
     this.episodicEffects.sort(this.compareEffectiveStartDates);
 
     // Current episodic effect (even null) are the first item in the arrays and open by default
-    ////this.episodicEffects.unshift(currentEpisodicEffect);
     this.status.isEpisodicEffectOpen.unshift(true);
```
oehm-smith: You used push() in one place and unshift() in another.  Lets be consistent.

lbodor: Which one should we use?

PR #114, opened 2016-11-21
SiteInfo re-architect - base and HumiditySensor

oehm-smith:
* Simple Events mechanism with GeodesyEvent struct
passed parent to child and back
* Remove can be done on any item
* Open / Close Above/Below functional but there is
no test if the above / below is already closed / open.

TODO:
* Currently base classes for Sensors.  Prolly immediately suitable for Antennas and Receivers etc..
* Fix in existing components
* Write unit tests


src/client/app/sensor-humidity/HumiditySensor.ts
```diff
@@ -0,0 +1,85 @@
+export interface BeginPosition {
+  value: string[];
+}
+
+export interface EndPosition {
+  value: string[];
+}
```
frankfu03: We may merge both BeginPosition, EndPosition and ValidTime2 interfaces into one. For example,

export interface Values {
    value: string[];
}

oehm-smith:
OK check it now.  Provide feedback and I'll **SQUASH** first **before merging**.

oehm-smith:
As per discussions in meeting - made fixes.  **Pleases review but dont' commit as I'll squash first.**
- HumidtySensors aren't open by default
- clicking on +/- or title opens / closes

src/client/app/sensor-humidity/humidity-sensors-group.component.ts
```diff
@@ -0,0 +1,283 @@
...
+      this.populateDefaultExistingItems(this.humiditySensorsOriginalProperties);
+      this.sortUsingComparator(this.humiditySensorsOriginalProperties);
+    }
+  }
+
+  getGeodesyEvent(): GeodesyEvent {
+    return this.geodesyEvent;
+  }
+
+  getComparator(): (obj1: HumiditySensorProperty, obj2: HumiditySensorProperty) => void {
```
lbodor: Eta-reduce. See https://en.wiktionary.org/wiki/eta_conversion.

oehm-smith:
I've got it about as small as possible (by moving as much as possible to abstract classes).  Check it!

src/client/app/shared/abstract-groups-items/AbstractGroup.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * A backup of the original list of items.  Used to diff against upon Save.
+   */
+  private itemOriginalProperties: any[];
+
+  /**
+   * Get the item name to be usedb in the subclasses.
+   */
+  abstract getWhatIsTheItemName(): string;
```
lbodor: Weird name. How about getItemName()?

src/client/app/shared/abstract-groups-items/AbstractGroup.ts
```diff
@@ -0,0 +1,269 @@
...
+  abstract makeItemsPropertyExist(itemsProperty: any);
+
+  /**
+   * Subclasses can create a comparator relevant for their data structures.  Reduce size in these by
+   * using getDateInstalled(), getBeginPositionDate.
+   *
+   * @param obj1
+   * @param obj2
+   */
+  abstract getComparator(obj1: any, obj2: any): number;
```
lbodor: This doesn't return a comparator. It is the compare funcation.

src/client/app/shared/abstract-groups-items/AbstractGroup.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  /**
+   * This is used in comparators but isn't a comparator - just a helper function.  In the comparator, extract the dates
+   * (using getDateInstalled(), getBeginPositionDate(), ...) and return compareDates(date1, date2)
+   * @param date1
+   * @param date2
+   * @return -1: date1 < date2; 1: date1 > date2; 0: date1 == date2
+   */
+  public compareDates(date1: string, date2: string): number {
```
lbodor: Good name! compareDates

src/client/app/shared/abstract-groups-items/AbstractGroup.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  /**
+   * Comparator for sorting by object.validTime.abstractTimePrimitive['gml:TimePeriod'].beginPosition.value[0].
+   * @param obj1
+   * @param obj2
+   * @returns {number}
+   */
+  comparatorEffectiveStartDates(obj1: any, obj2: any) {
```
lbodor: Prefer verbs for function names, not nouns.

src/client/app/site-info/site-info.component.ts
```diff
@@ -85,7 +86,7 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
       this.siteId = id;
     });
 
-    this.siteLogModel = {
+    this.siteLogModelXXX = {
```
lbodor: Bad name: what is XXX?

oehm-smith:
Fixed:
* SiteInfo opens on load
* HumiditySensors don't
* When the HS group is opened only the first one is showing
* Comparator 'this' problem fixed by replacing with statics method calls

Please review.  **Don't merge as I want to squash first.**

oehm-smith:
Fixed:
- Fixed setting 'Date Removed' on previous HS when create a new
- Fixed so HS Group opens when a new one is created

PR #119, opened 2016-11-25
added temperature sensors to site info screen

heyagosper:


heyagosper:
@oehm-smith could you review this PR for me? I don't think the Travis failure was related to my changes because I ran it again and it passed

PR #148, opened 2016-12-16
add template for text fields and some example use

heyagosper:
Three example fields added in pressure sensors
Manufacturer (required)
Serial Number (required)
Data Sampling Interval (not required)

there is functionality available that is not demonstrated in this PR: minimum length
for example: 
        <gnss-text-input 
              [model]="site.fourCharacterID"
              [name]="fourCharacterID"
              [label]="'Four Character Id'"
              [required]="'required'"
              [minlength]="'4'"
              [maxlength]="'9'">
         </gnss-text-input>

I think we would also want to pass in the units of measurement so that we can construct better validation messages (ie we don't necessarily want to use the label in those messages although it might be fine, I don't know)

This does not support type="number" fields

It uses template-driven forms and doesn't save THAT much lines of html code. But it is way neater and easier to scan.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "@angular/platform-browser": "2.0.2",
     "@angular/platform-browser-dynamic": "2.0.2",
     "@angular/router": "3.0.0",
-    "alertify.js": "^1.0.10",
+    "angular2-notifications": "0.4.41",
     "bootstrap": "3.3.7",
     "core-js": "^2.4.0",
     "es-module-loader": "^1.0.0",
--- a/src/client/app/pressure-sensor/pressure-sensor.component.html
+++ b/src/client/app/pressure-sensor/pressure-sensor.component.html
@@ -58,41 +58,35 @@
                     <div *ngIf=" pressureSensor == null ">
                         No information available for current GNSS pressure sensor
                     </div>
-                    <div *ngIf=" pressureSensor != null ">
-                        <div class="form-group">
-                            <label class="col-md-4 col-sm-4 col-xs-5 col-xxs-12 control-label required">Manufacturer</label>
-                            <div class="col-md-5 col-sm-6 col-xs-6 col-xxs-12">
-                                <input required type="text" maxlength="100" class="form-control"
-                                       [(ngModel)]="pressureSensor.manufacturer"
-                                       name="pressureSensorManufacturer_{{i}}"
-                                       #manufacturer="ngModel"/>
-                                 <small [hidden]="manufacturer.valid || manufacturer.pristine"
-                         			class="alert alert-danger">
-                    				"Manufacturer" is required
-                    			 </small>
-                            </div>
-                        </div>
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -135,7 +135,7 @@
     "@angular/platform-browser": "2.0.2",
     "@angular/platform-browser-dynamic": "2.0.2",
     "@angular/router": "3.0.0",
-    "alertify.js": "^1.0.10",
+    "angular2-notifications": "0.4.41",
```
oehm-smith: Is this from the old work where you tried to swap in ng2-notifications?

src/client/app/pressure-sensor/pressure-sensor.component.html
```diff
@@ -58,41 +58,35 @@
...
+                        <gnss-text-input
+                            [name]="'pressureSensorManufacturer'"
+                            [label]="'Manufacturer'"
+                            [(model)]="pressureSensor.manufacturer"
+                            [index]="[i]"
+                            [maxlength]="'100'"
+                            [required]="'required'">
+                        </gnss-text-input>
+
+                        <gnss-text-input
```
oehm-smith: How about trying to change to only two fields (model, descriptor (see next)).  This will reduce the boiler-plate and size of the htmls (further) and move the logic in to the `.ts` file.

````
<gnss-text-input2 # so can test out this idea first
  [descriptor]="descriptor"
  [(model)]="pressureSensor.manufacturer"></>

File: src/client/app/select-site/select-site.component.html [hunk 3]
```diff
@@ -1,3 +1,4 @@
+<simple-notifications [options]="notificationsOtions"></simple-notifications>
```
oehm-smith: old ng2-notifications?

src/client/app/select-site/select-site.component.ts
```diff
@@ -4,6 +4,7 @@ import { Subscription } from 'rxjs/Subscription';
 import { Subject } from 'rxjs/Subject';
 import 'rxjs/add/operator/debounceTime';
 import { DialogService, MiscUtilsService, CorsSiteService, ServiceWorkerService } from '../shared/index';
+import { NotificationsService } from 'angular2-notifications';
```
oehm-smith: old ng2-notifications?

src/client/app/shared/dynamic-form-fields/text-input.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  styleUrls: ['text-input.component.css']
+})
+export class TextInputComponent {
+  @Input() name: string = '';
+  @Input() label: string = '';
+  @Input() model: string = '';
+  @Input() index: string = '0';
+  @Input() required: string = '';
+  @Input() minlength: string = '';
+  @Input() maxlength: string = '';
```
oehm-smith: Should these be number type?

PR #149, opened 2016-12-16
Move duplicate getDate() method to MiscUtilsService

frankfu03:


src/client/app/shared/global/misc-utils.service.ts
```diff
@@ -13,6 +13,18 @@ export class MiscUtilsService {
   }
 
   /**
+   * Returns the date string (YYYY-MM-DD) from the date-time string (YYYY-MM-DDThh:mm:ssZ)
+   */
+  public getDate(datetime: string) {
```
lbodor: Specify return type please.

frankfu03:
Yes, all methods in MiscUtilsService should be declared as static ones. But we could not directly access to static methods in html.

frankfu03:
Can we merge this PR first? I will create a separate branch to make all methods from MiscUtilsService class to static ones.

PR #152, opened 2016-12-20
Fix tests after recent additions

oehm-smith:


karma.conf.js
```diff
@@ -75,6 +75,9 @@ module.exports = function (config) {
       { pattern: 'node_modules/foreach/**/*.js', included: false, watched: false },
       { pattern: 'node_modules/json-pointer/**/*.js', included: false, watched: false },
       { pattern: 'node_modules/scroll-into-view/**/*.js', included: false, watched: false },
+      { pattern: 'node_modules/raf/*.js', included: false, watched: false },
+      { pattern: 'node_modules/raf/**/*.js', included: false, watched: false },
```
lbodor: Do we need raf mentioned twice? Perhaps the second one is enough.

oehm-smith: Yep I took the first out and tests pass so I've pushed an updated version.

PR #153, opened 2016-12-21
Static utils methods

frankfu03:


src/client/app/shared/global/misc-utils.service.ts
```diff
@@ -3,19 +3,19 @@ import { Injectable } from '@angular/core';
 @Injectable()
 export class MiscUtilsService {
 
-  private _scrollIntoView: any = require('scroll-into-view');
+  private static _scrollIntoView: any = require('scroll-into-view');
```
lbodor: Typescript coding guidelines advise against prefixing private members with an underscore.

src/client/app/shared/global/misc-utils.service.ts
```diff
@@ -3,19 +3,19 @@ import { Injectable } from '@angular/core';
 @Injectable()
 export class MiscUtilsService {
```
frankfu03: Good suggestion! Relevant changes have been made as suggested.

PR #156, opened 2016-12-23
Gnss antenna

frankfu03:


frankfu03:
I will apply the new view model to Antenna soon.

PR #161, opened 2017-01-04
apply text input component to humidity sensor fields

heyagosper:


src/client/app/humidity-sensor/humidity-sensor-item.component.html
```diff
@@ -27,88 +27,90 @@
...
-            <div class="form-group">
-                <label class="col-md-4 col-sm-4 col-xs-5 col-xxs-12 control-label">Notes</label>
-                <div class="col-md-8 col-sm-8 col-xs-7 col-xxs-12">
-                    <textarea maxlength="2000" rows="2" class="form-textarea"
-                              [(ngModel)]="humiditySensor.notes"
-                              name="humiditySensorNotes_{{index}}"></textarea>
-                </div>
-            </div>
-        </div>
+
```
oehm-smith: What do you think about moving some of these fields to an options object / function in the component class?  I'll add some examples below.

src/client/app/humidity-sensor/humidity-sensor-item.component.html
```diff
@@ -27,88 +27,90 @@
...
+            <text-input
+              [name]="'humiditySensorManufacturer'"
+              [(model)]="humiditySensor.manufacturer"
+              [index]="[i]"
+              [label]="'Manufacturer'"
+              [maxlength]="100"
+              [required]="'required'">
+            </text-input>
+
```
oehm-smith:
```
<text-input
               [name]="'humiditySensorManufacturer'"
               [(model)]="humiditySensor.manufacturer"
               [index]="[i]" 
               [options]=getOptions('manufacturer')  # OR use an enum
/>

getOptions(): textOptions { # Typescript interface 
 ...
               return {label:"'Manufacturer'",
                          maxlength: "100",
                          required: "'required'"}
```

etc for the ones below.

I don't know if you can achieve the '[....]' when doing it this way.

PR #162, opened 2017-01-05
apply text input component to humidity sensor fields

heyagosper:


src/client/app/shared/dynamic-form-fields/form-field-configuration.service.ts
```diff
@@ -0,0 +1,34 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class FormFieldConfigurationService {
```
oehm-smith: I can't see where this is used?

oehm-smith: If you do do something like this (like my options I discussed in that closed pull request I believe) then put in the component.

PR #165, opened 2017-01-06
remove default calibration and start dates

heyagosper:
we should do this in all of the view model classes


oehm-smith:
Do it to all?  Only if that is what the business rules say.

PR #166, opened 2017-01-06
remove default dates from humidity sensor constructor

heyagosper:
and add to the newViewModelItem function
in humidity and pressure sensors
so that we get default dates in a new item but not when we load records
 with missing data from the database

### DIFF
--- a/src/client/app/humidity-sensor/humidity-sensor-view-model.ts
+++ b/src/client/app/humidity-sensor/humidity-sensor-view-model.ts
@@ -20,10 +20,9 @@ export class HumiditySensorViewModel extends AbstractViewModel {
 
   constructor() {
     super();
-    let presentDT: string = MiscUtils.getPresentDateTime();
 
-    this.startDate = presentDT;
-    this.calibrationDate = presentDT;
+    this.startDate = '';
+    this.calibrationDate = '';
     this.endDate = '';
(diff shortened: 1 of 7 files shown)


src/client/app/humidity-sensor/humidity-sensor-view-model.ts
```diff
@@ -20,10 +20,9 @@ export class HumiditySensorViewModel extends AbstractViewModel {
 
   constructor() {
     super();
-    let presentDT: string = MiscUtils.getPresentDateTime();
 
-    this.startDate = presentDT;
-    this.calibrationDate = presentDT;
+    this.startDate = '';
```
oehm-smith: Comment that default dates are set in 'new()'.

src/client/app/humidity-sensor/humidity-sensors-group.component.ts
```diff
@@ -48,6 +48,10 @@ export class HumiditySensorsGroupComponent extends AbstractGroup<HumiditySensorV
    * Other methods
    */
   newViewModelItem(): HumiditySensorViewModel {
-    return new HumiditySensorViewModel();
```
oehm-smith: Comment why setting default dates here rather than constructor.

PR #170, opened 2017-01-09
fix validation for required and invalid date strings

heyagosper:
I moved the logic into the component to debug it mainly but actually i think it is much nicer having it there rather than in the template.
Tested in the UI and it works very well now in Firefox and Chrome.
I haven't retrofitted this code into the old datetime component and I'm not interested in doing that, I'd prefer to replace them all with the new component.

### DIFF
--- a/src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
+++ b/src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
@@ -353,4 +353,20 @@ export class DatetimeInputComponent implements OnInit, DoCheck {
   public isRequired() : boolean {
     return this.required || (this.requiredIfNotCurrent && this.index[0] !== 0);
   }
+
+  /**
+   * Get a validation message to apply if the input string is:
+   *   1. required but not supplied, or
+   *   2. supplied but not a valid date
+   */
+  public getValidationMessage() {
+    let message = this.label + ' is ';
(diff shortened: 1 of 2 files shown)


src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
```diff
@@ -353,4 +353,20 @@ export class DatetimeInputComponent implements OnInit, DoCheck {
...
+   *   2. supplied but not a valid date
+   */
+  public getValidationMessage() {
+    let message = this.label + ' is ';
+    if (this.isRequired() && !this.datetimeDisplay) {
+      message += ' required ';
+    } else if (this.invalidDatetime) {
+      message += ' not valid ';
+    }
+    message += '(Valid Date Format: YYYY-MM-DD hh:mm:ss)';
```
oehm-smith: Is it possible to not hard-code this but have it come from the same place as the format?

heyagosper: it might be nice but since this pattern isn't actually used for the validation (we use a regular expression and not a pattern like DateFormatters in Java) it's would just be shifting deck chairs...
save the effort for some proper work on the validation framework?

oehm-smith: No worries.

PR #179, opened 2017-01-12
Remove array elements which are null

heyagosper:


lbodor:
Could you please rebase onto latest master?

PR #184, opened 2017-01-16
Apply view model to GNSS receiver component

frankfu03:


src/client/app/gnss-receiver/gnss-receiver-group.component.ts
```diff
@@ -0,0 +1,50 @@
...
+    super();
+  }
+
+  getItemName(): string {
+    return 'GNSS Receiver';
+  }
+
+  compare(obj1: GnssReceiverViewModel, obj2: GnssReceiverViewModel): number {
+    if (obj1 === null || obj2 === null) {
+      return 0;
```
lbodor: Doesn't zero mean the objects are equal? Function compareDates() below returns -1, 0, or 1, depending on whether the answer is respectively less than, equal, or greater than.

frankfu03: Yes, zero means the two objects are equal. The compare() method is used to sort a list of items in a descending date order. In this case, if either of the objects is null, there will be no sorting for them.

frankfu03: Yes, the returned value zero means the two objects are equal. The compare() method is used to sort a list of items in a descending date order. In this case, if either of the objects is null, there will be no change in order for them.

lbodor:
Rebase to resolve conflicts.

src/client/app/gnss-receiver/gnss-receiver-group.component.ts
lbodor: OK, for now. I've added a todo in humidity sensors to fix this. I don't think it's right for compare to return zero when its arguments are not equal. We will try to implement correct sorting and a correct compare function.

oehm-smith: I made the change to remove these lines in HumiditySensors and have been using it for a couple days without any errors (related to this).  

You might as well remove these lines.

lbodor: That's great news, let's remove them. I think we should turn on strict --strictNullChecks. See: https://blog.mariusschulz.com/2016/09/27/typescript-2-0-non-nullable-types.

PR #186, opened 2017-01-17
Implement deleted removed items

oehm-smith:
In the items, such as HumiditySensor-Item, the 'remove' button should be:
'remove' for new items that have been created, and
'delete' for existing ones and when this is clicked a warning dialog should be presented.

h2. Other Geodesy sensors items changes

Do this in the other `-items.html`:
* Change `<span class="hidden-xs">Remove</span>` to `<span class="hidden-xs">{{getRemoveOrDeletedText()}}</span>`


lbodor:
👍

heyagosper:
wouldn't it be easier for one dev to do them all?

PR #189, opened 2017-01-18
Improve remove delete items

oehm-smith:
Fulfilled https://gaautobots.atlassian.net/browse/GEOD-316:

* When create an item set the Change Tracking DateInserted field (actually was already doing that).
* When delete a just-created one it can be deleted (no record).
* When delete an existing:
** set the dateDeleted (actually was already doing that),
** put an entry into the deletedReason field
** DON'T DELETE FROM THE SITELOG - Keep the historical data.

As an additional extra I quickly (since the new architecture makes this easy) added a link to 'show deleted' along with the normally unseen fields.

* Lazar also suggested seeing if I can set the default fields in the constructor of the super class rather than require each component to do so.


src/client/app/shared/json-data-view-model/view-model/site-log-view-model.ts
```diff
@@ -14,6 +14,9 @@ export class SiteLogViewModel {
 }
 
 export class ViewSiteLog {
+    // activate this to remove deleted items from the SiteLog
+    readonly deleteEnabled: boolean = false;
```
lbodor: If this is always false, then we don't need it, so we should probably remove it.

oehm-smith: I considered that we might want to enable so leaving ourselves the option.  If the concensus is a definite no then I'll remove all that plumbing.  What do you think - definitely don't want?

lbodor: Let's delete it if don't need it, and add it when we do need it. is there any other plumbing to delete besides the boolean flag and the if-statement?

oehm-smith:
Updated code to remove any save functionality.  Tests all passing.

lbodor:
:+1:

PR #192, opened 2017-01-19
Extend delete/remove functionality to more components

heyagosper:


src/client/app/shared/abstract-groups-items/abstract-item.ts
```diff
@@ -79,21 +94,37 @@ export abstract class AbstractItem implements DoCheck, OnInit {
...
+          this.dialogService.confirmDeleteDialog(
+            this.getItemName(),
+            (deleteReason : string) => {
+               // ok callback
+               this.deleteItem(index, deleteReason);
+            },
+            function() {
+              // cancel callback
+              console.log('delete cancelled by user');
             }
```
lbodor: Prefer fat arrows, even when there is no reference to `this`.

lbodor:
👍

PR #193, opened 2017-01-19
Diff table update

oehm-smith:
# Changes:
* All Items (eg. Humidity Sensor (2017-01-24 - 2017-01-25)) are part of the one table so all columns line up vertically
* The column titles come out initially as first row and then at the start of a new item if 10 differences have been printed.  10 is a guess at a good number to balance between mobile and desktop
* Exclude some rows from showing in the diff - dateDeleted, dateInserted, deletedReason
* New items only have two columns (ie. no 'Old Value's)
* Deleted items only have one entry - 'Deleted'

### DIFF
--- a/src/client/app/shared/global/json-diff.service.ts
+++ b/src/client/app/shared/global/json-diff.service.ts
@@ -8,368 +8,369 @@ import {JsonPointerService} from '../json-pointer/json-pointer.service';
 /**
  * JSON Differencing service using https://github.com/flitbit/diff.
  */
-export enum DiffType {Edited, New, Deleted, NewArrayItem, DeletedArrayItem};
+export enum DiffType {Edited, New, Deleted, NewArrayItem, DeletedArrayItem}
+;
 
 export class DiffItem {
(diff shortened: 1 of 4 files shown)


src/client/app/shared/global/json-diff.service.ts
```diff
@@ -255,21 +255,35 @@ export class JsonDiffService {
...
         return ident;
     }
 
+    /**
+     * Currently not all viewModels have been created and so their dates may be objects instead of a string
+     *
+     * @param date to check if a string or object
+     * @return {boolean} true if given date value is a string
+     */
+    private checkIfDate(date: any): boolean {
```
heyagosper: don't like the name of this function, how about isDate(arg)?

lbodor: I agree.

lbodor: I'd like to encourage more comments of this kind. Naming matters.

PR #194, opened 2017-01-19
Fix error that occurs if a data item is not present

heyagosper:
bug in line 151 - (let dataModel of dataModels) 

if the collection is undefined Javascript crashes. To reproduce delete all items of a type like humidity sensors in the database and try to load the sitelog again

### DIFF
--- a/src/client/app/shared/json-data-view-model/json-view-model.service.ts
+++ b/src/client/app/shared/json-data-view-model/json-view-model.service.ts
@@ -140,37 +140,41 @@ export class JsonViewModelService {
     }
 
     /* ***************************** Helper functions ***************************** */
+
     /**
-     * Translate data model to view model
-     * @param dataModels - array of data model items to convert
-     * @param viewModelInstance - used as a template to copy and populate.  And has methods used.
-     * @returns {AbstractViewModel[]} that is the super type of all view model types
-     */
+    * Translate data model to view model
+    * @param dataModels - array of data model items to convert
+    * @param viewModelInstance - used as a template to copy and populate.  And has methods used.
+    * @returns {AbstractViewModel[]} that is the super type of all view model types
+    */
     private dataToViewModel<T extends AbstractViewModel>(dataModels: any[], type: {new(): T ;}): T[] {
         let viewModels: T[] = [];
-        for (let dataModel of dataModels) {
-            let newViewModel: T = new type();
-            let fieldMappings: FieldMaps = newViewModel.getFieldMaps();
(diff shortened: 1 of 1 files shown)


src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -139,38 +139,41 @@ export class JsonViewModelService {
...
+    if (viewModels) {
+      for (let viewModel of viewModels) {
+        let fieldMappings: FieldMaps = viewModel.getFieldMaps();
+        let dataModel: any = {};
+        DataViewTranslatorService.translateV2D(viewModel, dataModel, fieldMappings);
+        dataModels.push(dataModel);
+      }
     }
+  }
+
```
lbodor: Indent by 4 spaces please.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -139,38 +139,41 @@ export class JsonViewModelService {
...
+  /* ***************************** Helper functions ***************************** */
+  /**
+   * Translate data model to view model
+   * @param dataModels - array of data model items to convert
+   * @param viewModelInstance - used as a template to copy and populate.  And has methods used.
+   * @returns {AbstractViewModel[]} that is the super type of all view model types
+   */
+  private dataToViewModel<T extends AbstractViewModel>(dataModels: any[], type: {new(): T ;}): T[] {
+    let viewModels: T[] = [];
+    if (dataModels) {
```
oehm-smith: No don't do it this way.  Instead of 

```
    public dataModelToViewModelJson(dataModelJson: any): SiteLogViewModel {
        let siteLogDataModel: SiteLogDataModel = <SiteLogDataModel> dataModelJson;
```

With SiteLogDataModel as an interface, make it a class taking a dataModelJson and build empty arrays/objects if no data available.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -140,37 +140,41 @@ export class JsonViewModelService {
...
-            dataModels.push(dataModel);
+        if (viewModels) {
+            for (let viewModel of viewModels) {
+                let fieldMappings: FieldMaps = viewModel.getFieldMaps();
+                let dataModel: any = {};
+                DataViewTranslatorService.translateV2D(viewModel, dataModel, fieldMappings);
+                dataModels.push(dataModel);
+            }
         }
-        return dataModels;
```
oehm-smith: These are needed - tests are failing (compile error).

PR #195, opened 2017-01-19
GEOD-315 - Cancelled new items should delete them

oehm-smith:
We had a bug where:
* add item
* delete item
* save and shown as a diff (ie. still existed)

This made sure such items are deleted.


lbodor:
👍

PR #196, opened 2017-01-20
Change data site log to class

heyagosper:


src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
```diff
@@ -1,33 +1,77 @@
...
+     */
+    private getSuppliedValueOrEmptyValue(value : any, isArray : boolean) : any {
+        if (value !== null && value !== undefined) {
+           return value;
+        } else {
+           return isArray ? [] : new Object();
+        }
+    }
 
-export interface SiteLogDataModel {
```
oehm-smith: Ok to getting rid of this extra level.  Hope you found all areas it was used!

src/client/app/shared/json-data-view-model/json-view-model.service.spec.ts
```diff
@@ -82,48 +83,30 @@ export function main() {
...
-      // expect(siteLog.gnssReceivers.length).not.toBe(0);
-      // expect(siteLog.gnssAntennas).toBeDefined();
-      // expect(siteLog.gnssAntennas.length).not.toBe(0);
-      // expect(siteLog.surveyedLocalTies).toBeDefined();
-      // expect(siteLog.surveyedLocalTies.length).not.toBe(0);
-      // expect(siteLog.frequencyStandards).toBeDefined();
-      // expect(siteLog.frequencyStandards.length).not.toBe(0);
+      let siteLogDataModel: SiteLogDataModel = jsonViewModelService.viewModelToDataModelJson(siteLogViewModel);
 
       // TODO - introduce HumiditySensorProeprtyType intermediatery
```
oehm-smith: Handled internally.  You can lose this comment also.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -25,57 +24,36 @@ export class JsonViewModelService {
...
 
-        // siteLogViewModel.siteIdentification = SiteIdentification.translateDataToView(siteLogDataModel);
-        // siteLogViewModel.siteLocation = SiteLocation.translateDataToView(siteLogDataModel);
-        // siteLogViewModel.siteIdentification = SiteIdentification.translateDataToView(siteLogDataModel);
-        // siteLogViewModel.siteLocation = SiteLocation.translateDataToView(siteLogDataModel);
-        // siteLogViewModel.gnssReceivers = GnssReceivers.translateDataToView(siteLogDataModel);
-        siteLogViewModel.siteLog.gnssAntennas = this.dataToViewModel(siteLogDataModel['geo:siteLog'].gnssAntennas, GnssAntennaViewModel);
-        // siteLogViewModel.surveyedLocalTies = SurveyedLocalTies.translateDataToView(siteLogDataModel);
-        siteLogViewModel.siteLog.frequencyStandards = this.dataToViewModel(siteLogDataModel['geo:siteLog'].frequencyStandards,
+        siteLogViewModel.siteLog.gnssAntennas = this.dataToViewModel(siteLogDataModel.gnssAntennas, GnssAntennaViewModel);
```
oehm-smith: Yes it looks so much better without the `['geo:siteLog']` indirection.

src/client/app/shared/jsonix/jsonix.service.ts
```diff
@@ -67,23 +67,15 @@ export class JsonixService {
      * @param json that is 'valid' to translate
      * @returns {string} the valid GeodesyMl
      */
-    jsonToGeodesyML(json: any): string {
-        let jsonObj: string = '';
+    jsonToGeodesyML(jsonObj: Object): string {
```
oehm-smith: I don't think this change is correct.  `any` is the "new" `Object`.  What you have prolly transpiles the same though it would be better to keep the `any`.  Did something fail without this change?

heyagosper: new plan: don't let callers pass a string into this method, only a valid JSON object. While using any will work, using Object gives callers a useful hint. 
I'll update the doc comment on this method.

src/client/app/shared/json-data-view-model/json-view-model.service.spec.ts
heyagosper: ok. i uncommented all those extra tests as well, it was that or delete them

src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
heyagosper: yes. I think I did. Because actually I got rid of the internal type and moved the data into the higher level type.

src/client/app/shared/json-data-view-model/data-view-translator.spec.ts
```diff
@@ -1,11 +1,11 @@
+import {SiteLogDataModel} from './data-model/site-log-data-model';
 import {DataViewTranslatorService} from './data-view-translator';
 import {JsonViewModelServiceSpecData} from './json-view-model.service.spec.data';
 import {FieldMaps} from './field-maps';
-import {SiteLogDataModel} from './data-model/site-log-data-model';
 import {HumiditySensorViewModel} from '../../humidity-sensor/humidity-sensor-view-model';
 
 export function main() {
-  let completeValidSitelog: SiteLogDataModel = JsonViewModelServiceSpecData.data();
+  let completeValidSitelog: any = JsonViewModelServiceSpecData.data();
```
heyagosper: Because in the real world the data is now a JSON object and is not a SiteLogDataModel. I think the spec.data should reflect this. It's up to the caller to use the data in whatever way they want, in this case to construct a data model object.
I think a more interesting question is, why do we even need a SiteLogDataModel object? It feels like an unnecessary layer of mapping between JSON->View Model

src/client/app/shared/json-data-view-model/json-view-model.service.spec.data.ts
```diff
@@ -1,7 +1,6 @@
-import {SiteLogDataModel} from './data-model/site-log-data-model';
 export class JsonViewModelServiceSpecData {
-    public static data(): SiteLogDataModel {
-        return <SiteLogDataModel>{
+    public static data(): any {
```
heyagosper: see my other comment: the data is a JSON object not an egeodesy object

src/client/app/shared/json-data-view-model/data-view-translator.spec.ts
oehm-smith: We use SLDM as we expect data in this format.  We went with typescript so that we could type data to avoid us having to guess and / or test.  By trying to new SLDM() with data we know immediately if there is a problem if it isn't the correct "shape".

src/client/app/shared/json-data-view-model/json-view-model.service.spec.data.ts
oehm-smith: But it is as we are using objects that are the correct "shape".

src/client/app/shared/jsonix/jsonix.service.ts
oehm-smith: Yep Object is a good idea (http://stackoverflow.com/questions/18961203/typescript-any-vs-object)

PR #199, opened 2017-01-23
GEOD-320 clean up site info component

heyagosper:


src/client/app/site-info/site-info.component.ts
```diff
@@ -103,21 +83,11 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
...
       gnssReceivers: [],
       surveyedLocalTies: [],
-      localEpisodicEvents: [],
-      humiditySensors: [],
-      pressureSensors: [],
-      temperatureSensors: [],
-      waterVaporSensors: []
     };
 
     this.siteLogOrigin = {
```
oehm-smith: I don't think this is needed either, as long as it is declared.  Could you remove and try again.

src/client/app/site-info/site-info.component.ts
```diff
@@ -180,11 +141,6 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
           this.siteLogModel = {
```
oehm-smith: I don't believe this is needed either, as long as it is declared.  Could you remove and test again.

src/client/app/site-info/site-info.component.ts
heyagosper: tests and UI appear to still work fine

lbodor:
:+1:

PR #203, opened 2017-01-24
GEOD-322 Use value rather than codeTypeValue

heyagosper:


lbodor:
👍 

I agree that `value` is correct. Here is some related information. The target XML type in GeodesyML for receiverType is a subtype of http://www.datypic.com/sc/niem30/t-gco_CodeListValue_Type.html. Currently, our sample GeodesyML sitelogs in project gws-system-test look like this:
```xml
<geo:receiverType
    codeList="http://xml.gov.au/icsm/geodesyml/codelists/antenna-receiver-codelists.xml#GeodesyML_GNSSReceiverTypeCode"
    codeListValue="AOA ICS-4000Z"
    codeSpace="urn:xml-gov-au:icsm:egeodesy:0.3">
AOA ICS-4000Z
</geo:receiverType>
```
I think that codeListValue should be a full URI `http://xml.gov.au/icsm/geodesyml/codelists/antenna-receiver-codelists.xml#ReceiverTypeCode_AOA_ICS-4000Z`.

I also think that the GNSS Site Manager should generate GeodesyML documents with all attributes of `gco:CodeListValue_Type`s.

PR #205, opened 2017-02-01
Auth

oehm-smith:
Basic login with login/logout link, username and disabled fields.  Fit for demonstration.
    
* Disabled fields are only those in SiteInfo as for others need to modify components


oehm-smith:
Error fixed - build should pass now.

oehm-smith:
Everyone happy with this?

src/client/app/shared/global/user-auth.service.ts
```diff
@@ -0,0 +1,212 @@
+import {Injectable} from '@angular/core';
+import {ConstantsService} from './constants.service';
+
+/**
+ * Service for authentication and authorisation of subjects.
+ */
+
+export class IdToken {
+    private _firstName: string; // Geodesy Extension
```
lbodor: Let's not start using underscores for private properties: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines.

oehm-smith: I can't find anything definitive about the proper coding standard when it comes to TS properties (with get and set).  That page you link to says "Do not use "_" as a prefix for private properties." but not what to use.  There doesn't seem to be a lot of opinion about this.  I found http://stackoverflow.com/questions/21244855/whats-the-standard-naming-convention-for-a-propertys-backing-field from 2 years ago.  A comment on that page says:

> @Sam the TypeScript team announced pretty early on that they wanted to see what the community did before recommending this kind of thing. Although the MS C# standard states that you shouldn't use an _ to prefix private fields, it is becoming increasingly accepted that it is better to do that than rely on prefixing a variable with this. when there is a conflict. I think it is likely that the underscore convention will become the accepted standard as Basarat says - but don't expect MS to dictate it for you at this stage. – Sohnee Jan 21 '14 at 11:21

This is a blog at Microsoft - http://blogs.microsoft.co.il/gilf/2013/01/22/creating-properties-in-typescript/ and he is using _.

It seems the community is going with '_' prefixes for properties.  We just chatted and you have agreed (you read https://github.com/angular/angular.io/issues/1108 - see comments at end).

lbodor: Yes, I agree.

oehm-smith:
I've removed the login html page now.  You can set the template to ''.

lbodor:
Superceded by #209

PR #219, opened 2017-02-27
IntelliJ Code Format for Typescript

oehm-smith:
* Incluces spaces inside import braces eg { abc }.
* Devs will need to import this


heyagosper:
I'm using eclipse. will someone buy me IDEA?

lbodor:
:+1:

PR #226, opened 2017-03-05
Disable site info input fields if unauthenticated

lbodor:


oehm-smith:
Yep that is cleaner. :+1:

PR #230, opened 2017-03-08
Confirm close

frankfu03:


src/client/app/app.component.ts
```diff
@@ -57,13 +60,41 @@ export class AppComponent {
...
   onClose(event: any) {
     if (event) {
-      this.siteInfo.goBack();
+      this.siteInfo.close();
+    }
+  }
+
+  /**
+   * Navigate to the "Search/select Sites" page triggered by the "Select Site" button on Toolbar component.
+   * Prompt users whether to abort the action if changes have been made on the Site Log Info page.
```
oehm-smith: Where's the prompt?

frankfu03: It is in the close() method of SiteInfo component.

PR #236, opened 2017-03-15
Check user's authorities on loading SiteInfo page

frankfu03:


src/client/app/site-info/site-info.component.ts
```diff
@@ -224,7 +230,34 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
...
+  private checkUserAuthorities() {
+    if (!this.siteId) {
+        this.hasEditRole = false;
+        return;
+    }
+
+    let user: User = this.userAuthService.getUser();
+    if ( !user || !user.profile || !user.profile.authorities ) {
+        this.hasEditRole = false;
+    } else if (user.profile.sub.toLowerCase() === 'amadmin'
```
oehm-smith: Do we really want this since that means the app is controlling auth.  Eg. what if one of those changes then this code will need to also.  I would drop the 'amadmin' test.  Discuss!

src/client/app/site-info/site-info.component.ts
```diff
@@ -224,7 +230,34 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
...
+        this.hasEditRole = true;
+    } else {
+        let myAuthority: string = 'edit-' + this.siteId.toLowerCase();
+        for (let authority of user.profile.authorities) {
+            if (myAuthority === authority.toLowerCase()) {
+                this.hasEditRole = true;
+            }
+        }
+    }
+    console.log('@@@@@@@@@@@@ hasEditRole='+this.hasEditRole);
```
oehm-smith: Maybe just '<userX> has editrole= ... (i guess this is debug!)

src/client/app/site-info/site-info.component.ts
lbodor: I agree that we should drop the check the particular user. For testing let's just create a user.x that belongs to group supersuser.

src/client/app/site-info/site-info.component.ts
```diff
@@ -224,7 +230,34 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
     this.siteLogOrigin = MiscUtils.cloneJsonObj(this.siteLogModel);
   }
 
-  isUserLoggedIn(): boolean {
-    return this.userAuthService.getUser() !== null;
+  private setupAuthSubscription() {
```
oehm-smith: I'd move this to UserAuthService and call out / get Subscription.

src/client/app/site-info/site-info.component.ts
frankfu03: This condition (user.profile.sub.toLowerCase() === 'amadmin') is for testing and has been removed.

src/client/app/site-info/site-info.component.ts
frankfu03: yeah, removed.

src/client/app/site-info/site-info.component.ts
frankfu03: we may discuss it later as I need more explanation.

frankfu03: The checkUserAuthorities() function has been moved to UserAuthService as suggested.

PR #241, opened 2017-03-21
Add GA standard footer to the bottom of web page

frankfu03:


lbodor:
:+1:

src/client/app/app.component.html
```diff
@@ -2,3 +2,4 @@
 <div autoHeight>
     <router-outlet (activate)="onActivate($event)"></router-outlet>
 </div>
+<sd-footer></sd-footer>
```
oehm-smith: Would you be able to try and get this fitting within SiteInfo form?  Since would be better to have the toolbar buttons within the form.  The footer would be a good first cut since it is much easier.

PR #244, opened 2017-03-29
Add user registration form

lbodor:


src/client/app/shared/global/user-auth.service.spec.ts
```diff
@@ -1,39 +0,0 @@
-import { UserAuthService } from './user-auth.service';
```
oehm-smith: We should fill this out.

src/client/app/user-registration/user-registration.component.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+                            'Lazar',
+                            'Bodor',
+                            'Geoscience Australia',
+                            'Software developer',
+                            'lazar.bodor@ga.gov.au',
+                            '0451061798',
+                            'ADE1',
+                        );
+
+                        const setValue = function(value: string, formControlName: string): void {
```
oehm-smith: You might prefer the form:

    const setValue = (value:string, formControlName: string): void => { ... }

src/client/app/user-registration/user-registration.component.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+import { RouterTestingModule } from '@angular/router/testing';
+import { BaseRequestOptions, ConnectionBackend, Http, HttpModule } from '@angular/http';
+import { MockBackend } from '@angular/http/testing';
+import { UserRegistrationComponent } from './user-registration.component';
+import { UserRegistrationModule } from './user-registration.module';
+import { UserAuthService, UserRegistration } from '../shared/global/user-auth.service';
+import { DialogService } from  '../shared/global/dialog.service';
+import { Observable } from 'rxjs/Observable';
+import { Subscriber } from 'rxjs';
+import * as _ from 'lodash';
```
oehm-smith: Didn't you want:

    import * as lodash from 'lodash';

lbodor: I didn't now that underscore was the adopted convention for lodash. Now I do.

src/client/app/user-registration/user-registration.component.ts
```diff
@@ -0,0 +1,47 @@
...
+import { UserAuthService } from '../shared/global/user-auth.service';
+
+@Component({
+    moduleId: module.id,
+    selector: 'sd-user-registration',
+    templateUrl: 'user-registration.component.html',
+})
+export class UserRegistrationComponent {
+
+    userRegistrationForm = new FormGroup({
```
oehm-smith: Why not use the builder:



    uRF = formBuilder.group({...});

Like I thought we decided on using?

src/client/app/shared/global/user-auth.service.ts
```diff
@@ -1,7 +1,21 @@
...
+    constructor(
+        readonly firstName: string,
+        readonly lastName: string,
+        readonly organisation: string,
+        readonly position: string,
+        readonly email : string,
+        readonly phone : string,
+        readonly remarks: string,
+      ) {}
+}
```
frankfu03: Should user provide country/state (or even address) as well?

lbodor: Probably, but we can add those later. This is a start.

PR #245, opened 2017-03-29
Set CHROME_BIN environment variable

lbodor:


oehm-smith:
:+1:  :french_guiana:

PR #255, opened 2017-04-13
Dirty fields and json diff

frankfu03:


src/client/app/shared/abstract-groups-items/abstract-group.ts
```diff
@@ -3,6 +3,9 @@ import { AbstractViewModel } from '../json-data-view-model/view-model/abstract-v
 import * as lodash from 'lodash';
 
 export abstract class AbstractGroup<T extends AbstractViewModel> {
+    NONE: number = 0;
```
oehm-smith: What do you think about:
* Moving these 'constants' to an enum
* Change the method in the template from {{getValidationCode() == MODIFIED}} to {{isModified()}}

I believe this is better as it moves logic out of the templates and makes it easier to change.

src/client/app/shared/abstract-groups-items/abstract-group.ts
```diff
@@ -96,6 +99,21 @@ export abstract class AbstractGroup<T extends AbstractViewModel> {
         return this.hasGroupANewItem;
     }
 
+    getValidationCode(): number {
+        if (this.isEmptyCollection()) {
+            return 0;
```
oehm-smith: Do you mean 'NONE'? :)

src/client/app/shared/abstract-groups-items/abstract-group.ts
```diff
@@ -118,7 +136,7 @@ export abstract class AbstractGroup<T extends AbstractViewModel> {
     }
 
     isEmptyCollection(): boolean {
-        return (this.itemProperties === null || this.itemProperties.length === 0);
+        return (this.itemProperties === undefined || this.itemProperties === null || this.itemProperties.length === 0);
```
oehm-smith: This should be:

{{return (! this.itemProperties || this.itemProperties.length === 0);}}

src/client/app/shared/abstract-groups-items/abstract-group.ts
```diff
@@ -96,6 +99,21 @@ export abstract class AbstractGroup<T extends AbstractViewModel> {
         return this.hasGroupANewItem;
     }
 
+    getValidationCode(): number {
+        if (this.isEmptyCollection()) {
+            return 0;
+        }
+        let hasChanges: boolean = false;
+        for (let item of this.itemProperties) {
```
oehm-smith: And what you should do instead of this is move to working off the FormGroup.  We'll look at this after my code is merged.  And use a subscription or similar so learn of changes immediately without having to re-evaluate all the time (I think reevaluation must be going on).  Something for us to look at.

src/client/app/shared/global/json-diff.service.ts
```diff
@@ -220,53 +220,57 @@ export class JsonDiffService {
         let itemsPrinted: number = 0;
         let itemsBeforePrintHeader: number = 10;
 
-        let tableHtml: string = '<table class="table table-striped table-hover">\n';
+        let tableHtml: string = '<ol>';
```
oehm-smith: Wanna stop calling a `table` now that its list(s)? :face_with_head_bandage:

src/client/app/shared/global/json-diff.service.ts
```diff
@@ -220,53 +220,57 @@ export class JsonDiffService {
...
+                continue;
+            }
 
-            tableHtml += '<thead><tr><th colspan="3">&nbsp</th></tr>\n';
-            tableHtml += '<tr><th colspan="3">' + itemHeader + '</th></tr>\n';
-            if (this.isDeletedDiff(diffItems)) {
-                tableHtml += '</thead>\n<tbody>\n';
-                tableHtml += '<tr><td colspan="3">Deleted</td></tr></tbody>\n';
+            itemHeader = this.extractItemHeader(mapKey) + this.getExtraHeaderLabel(diffItems[0].diffType);
+            tableHtml += '<li class="table-caption">' + itemHeader + '</li>\n';
```
oehm-smith: And no longer a `table-caption`.

src/client/app/shared/global/json-diff.service.ts
```diff
@@ -220,53 +220,57 @@ export class JsonDiffService {
...
+            tableHtml += '<table class="table table-striped table-hover">\n';
```
oehm-smith: Oh sorry, I see it still is made of tables.  Still maybe not call `tableHtml` but `diffHtml`.

oehm-smith:
And does the `JsonDiffService.spec` still run successfully after your changes?

src/client/app/shared/abstract-groups-items/abstract-group.ts
frankfu03: yeah. Changed.

src/client/app/shared/abstract-groups-items/abstract-group.ts
frankfu03: Yeah, it is done.

src/client/app/shared/abstract-groups-items/abstract-group.ts
frankfu03: OK.

src/client/app/shared/global/json-diff.service.ts
frankfu03: Yeah, I was thinking of that, but prefer not to do too much changes.

src/client/app/shared/global/json-diff.service.spec.ts
```diff
@@ -811,7 +811,7 @@ export function main() {
       expect(diffItem.oldValue).toEqual('');
       expect(diffItem.newValue).toBeDefined();
       let humiditySensor: HumiditySensorViewModel = <HumiditySensorViewModel> diffItem.newValue.humiditySensor;
-      expect(diffItem.identifier).toEqual('New - ' + MiscUtils.prettyFormatDateTime(humiditySensor.startDate));
+      expect(diffItem.identifier).toEqual(MiscUtils.getDate(humiditySensor.startDate));
```
oehm-smith: I guess you changed it from saying 'New <date>'.  You should test that it is 'new'.

frankfu03: The diffTypes for new items have been already tested:
expect(diffItem.diffType).toEqual(DiffType.NewArrayItem);

src/client/app/frequency-standard/frequency-standard-group.component.html
```diff
@@ -1,4 +1,5 @@
-<div class="panel panel-level-1">
+<div class="panel panel-level-1"
+     [ngClass]="{'panel-dirty': getValidationCode() == MODIFIED, 'panel-invalid': getValidationCode() == INVALID}">
```
oehm-smith: Frank and I just discussed and I realised that we don't need this code (in the HTML like here or in the .ts).  The Reactive Forms wills set 'ng-dirty' or 'ng-invalid' etc....  Though I do need to add `[formGroup]="siteInfoForm"` to the GROUP components (which will take 2minutes).  The JSON diff and css mods are still  needed.

src/client/app/shared/abstract-groups-items/abstract-group.ts
```diff
@@ -118,7 +136,7 @@ export abstract class AbstractGroup<T extends AbstractViewModel> {
     }
 
     isEmptyCollection(): boolean {
-        return (this.itemProperties === null || this.itemProperties.length === 0);
+        return (!this.itemProperties || this.itemProperties.length === 0);
```
oehm-smith: But keep this change.

src/client/app/shared/abstract-groups-items/abstract-item.ts
```diff
@@ -4,8 +4,8 @@ import { DialogService } from '../index';
 
 export abstract class AbstractItem implements DoCheck, OnInit {
     protected isNew: boolean = false;
-
     protected isOpen: boolean = false;
+    protected hasErrors: boolean = false;
```
oehm-smith: Won't need this anymore.

src/client/app/shared/global/json-diff.service.ts
```diff
@@ -169,10 +169,10 @@ export class JsonDiffService {
     }
```
oehm-smith: Keep all these changes.

src/client/app/shared/global/json-diff.service.ts
oehm-smith: Lets keep it semantically correct since variable names help explain functionality.

frankfu03: OK. It has been changed to diffHtml.

src/client/app/frequency-standard/frequency-standard-item.component.html
```diff
@@ -21,7 +21,7 @@
     </span>
     <span *ngIf="frequencyStandard.dateDeleted" class="delete-marker"
         title="It will be deleted when saving all changes for the site">
-        (Marked to delete)
+        (<span [ngClass]="{'hidden-xxxs': index != 0}">marked </span>to delete)
```
oehm-smith: Explain to me the purpose of this.  I'm reading it as - if the item isn't the first then the 'hidden-xxxs' class will be applied so that 'marked ' is not seen on small screens.  Why the guard on index?

frankfu03: If it is not the current Frequency Standard (index!=0) which has start/end date range on header, hide the word "marked" to short the header. For the current Frequency Standard (index==0) which has only the start date on header, show the "marked" on the header.

PR #263, opened 2017-04-24
Reactive model g379 reformat

oehm-smith:


oehm-smith:
Frank verbally said he's happy with this.

PR #266, opened 2017-04-24
GEOD-379 - Fix dirty and valid so only parents view changes

oehm-smith:
* ng-dirty is applied to modified form fields, its form and any parent forms.
* There was a bug where it seems to also apply to the forms with the same form variable name.  So all sibling forms in our case.
* Fixed by working around and not using 'ng-dirty' and 'ng-invalid' but our own 'gnss-dirty' and 'gnss-invalid'


frankfu03:
The working-around works fine. But items' headers are not set as dirty when adding new items or deleting existing one. In addition, the SiteInfo header is marked as dirty when changes are made in other groups (Receivers, Antennas). I think that SiteInfo component is a sibling of all other groups.

oehm-smith:
I fixed from feedback Frank gave and will merge this.

PR #267, opened 2017-04-26
GEOD-379 - Due to modified SiteInfo move obj into place

oehm-smith:
* For the recent FixDirty commit, the SiteLocation, SiteIdentification
and all the SiteContacts were moved under an extra form so that we
could see when they are dirty (I termed this the 'artificial form'
since it isn't in the data, yet the data structure gets represented
by it).
* And ofcourse these components are in the root of the
SiteLogModel.  This change moves those components back to the root.
* Necessary for the diff and save


src/client/app/site-info/site-info.component.ts
```diff
@@ -377,4 +378,26 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
             }
         }
     }
+
+    /**
+     * Due to how we artifically nest SiteLocation and SiteIdentification under a SiteIdentificationForm, this data is not in the
```
frankfu03: SiteInformationForm?

frankfu03:
Looks good. :+1:

PR #268, opened 2017-04-27
Reactive model g379 bug form holding empty values

oehm-smith:
Changed to doing 'lazy loaded' forms by not creating the form controls until the Group is opened.

The reason for this is because prior to this change, the forms with all controls were being created and which didn't get populated until the Group was opened.  So when you saved, the blank default fields from all the forms under Groups that hadn't been opened were updating the legitimate values in the SiteLogModel.


frankfu03:
All good. :+1:

PR #269, opened 2017-04-27
Remove home directory

heyagosper:


oehm-smith:
Comments:
* We have an About and App components - do the equivalent e2e tests eg.  `src/e2e/specs/about.component.e2e-spec.ts` run successfully?
* Generally I wouldn't throw all the e2e's away as it will give you something to work from

frankfu03:
Actually, the directories of "home", "shared/name-list", "shared/navbar" were from the seed project and are not used in our project. We may delete all of them.

oehm-smith:
Given your comments about you will go find those files again in the seed when you want something to work from.  Then after deleting the other files as suggested by frank :+1:

PR #271, opened 2017-04-28
GEOD-370 - Dis/En able save & revert button & link

oehm-smith:
* Also removed 'is modified' debug and changed that to highlight
the SiteId instead if dirty.


frankfu03:
:+1:

PR #303, opened 2017-05-08
Declare return types

lbodor:


src/client/app/shared/abstract-groups-items/abstract-item.ts
```diff
@@ -176,7 +176,7 @@ export abstract class AbstractItem implements OnInit, OnChanges {
     /**
      * Patching (or setting) is used to apply the values in the model to the form.
      */
-    protected patchForm() {
+    protected patchForm(): void {
```
oehm-smith: void is inferred as default

PR #313, opened 2017-05-09
Responsible parties

frankfu03:


src/client/app/responsible-party/responsible-party-group.component.ts
```diff
@@ -46,24 +51,39 @@ export class ResponsiblePartyGroupComponent extends AbstractGroupComponent<Respo
...
+            this.setItemsCollection(siteLogModel[this.partyType.getObjectName()]);
+            this.init();
+        }
+    }
+
+    @Input()
+    set originalSiteLogModel(originalSiteLogModel: any) {
+        if (originalSiteLogModel) {
+            this.setItemsOriginalCollection(originalSiteLogModel[this.partyType.getObjectName()]);
+        }
```
lbodor: The two inputs above (siteLogModel and originalSiteLogModel) have been moved to AbstractGroupComponent earlier today. Could they stay there?

lbodor:
:+1:

src/client/app/responsible-party/responsible-party-group.component.ts
frankfu03: Yeah, the two inputs have been removed from responsible-party-group.component.ts.

But we have to keep them in site-identification and site-location components as the two components do not extend from AbstractGroupComponent.

frankfu03:
Yeah, the two inputs have been removed from responsible-party-group.component.ts.

But we have to keep them in site-identification and site-location components as the two components do not extend from AbstractGroupComponent.

PR #331, opened 2017-05-11
Use form dirty flag to decide to save or not

lbodor:


frankfu03:
:+1:

src/client/app/site-info/site-info.component.ts
```diff
@@ -177,9 +177,7 @@ export class SiteInfoComponent implements OnInit, OnDestroy {
          an inital copy was added to the SiteLogModel and SiteLogOrigin.  And in the form model too of course. */
         _.merge(this.siteLogModel, formValueClone);
 
-        let diffMsg: string = this.jsonDiffService.getJsonDiffHtml(this.siteLogOrigin, this.siteLogModel);
-
-        if (diffMsg === null || diffMsg.trim() === '') {
+        if (!this.isFormDirty()) {
```
oehm-smith: You should also check if `isValid()`

oehm-smith: It will be useful to have the information banner that says 'form dirty' or whatever so they know why the button is disabled (as per that other Issue we have open).  Oh yeah, we really want to have the `isValid()` on the button disabled attribute.

PR #333, opened 2017-05-11
Put contacts into the right place in the form

lbodor:


oehm-smith:
Closing this as there are too many date ties between the formModel and the viewModel.  This may be something we can revisit later on.

PR #338, opened 2017-05-12
This is not patching

lbodor:
The code in `patchForm` function does not perform any patching. Instead, it sets up the form controls, the dirty state, and sets the form values -- a perfect mix of work for the generic `ngOnInit`.


oehm-smith:
The reason I wrote it like this is to comment the code with a method name.   The method name succinctly describes what this code does. Extracting the code into `ngOnInit()` make that block just more code that you have to read to understand.  You could put a comment to say `// this code applies the data to the form` but then the trouble with comments (and why a method name is preferred) is because comments can become stale or dis-connected from the code.  

Uncle Bob (Robert C Martin) discusses this.

I don't approve this PR :).

lbodor:
There is nothing remotely resembling patching in this bit of code to motivate the name patchForm.

oehm-smith:
:+1:  but there is a conflict.

PR #343, opened 2017-05-15
added guard for field maps not existing

heyagosper:
I am still not sure why "getFieldMaps" is not a function... maybe the new record is not a fully-fledged view model at that stage (and when you cancel one it isn't removing it from the site log view model)

I'm not sure whether this is masking a code smell or not, it does prevent the error and conceptually seems ok to me:

"In the case where there is a view model that does not have mapped fields for some reason, don't try and create a data model out of it."


oehm-smith:
I'd rather work out why it was trying to call an object it thought was an AbstractViewModel.

PR #352, opened 2017-05-16
Date time bug g426 try2

oehm-smith:
Fixed bug where DateTime wasn't being set as dirty once a change was made.  Required updating to a FormControl for the new reactive forms.

* Incomplete work - I plan to do the other work on subsequent PR(s)
* DateTimes now a static text box (instead of text input) and the DateTimePicker widget
* Date picker ignores any timezone and assumes local.  It outputs local such (eg. GMT+10) - set in our view model via the `formControlName`.  But our dates need to be UTC. 
* ViewModel changed to a 'date' type from 'string' for the Dates.  They are still actually a string, but 'date' is used to indicate may need to change the format of the date.   This is required because the data from the DatePicker is like 'Tue 04 Sept ...' but we need in '20170904T0:00:000Z' format in the data.  The translate is done in the ViewToData translator.

## Todo in later PR(s)
* Fix TimeZones as above
* Time setting can't be done until this TimeZone problem is fixed
* OK and Cancel buttons currently only close the dialog.  Need to implement Cancel to restore previous date.
* Use Date type in the view model exclusively (no more strings - but translate to/from data model)

### DIFF
--- a/src/client/app/shared/dynamic-form-fields/abstract-gnss-controls.ts
+++ b/src/client/app/shared/dynamic-form-fields/abstract-gnss-controls.ts
@@ -37,7 +37,7 @@ export abstract class AbstractGnssControls {
         } else {
             dirty = control.dirty;
         }
-        console.warn('  isDirty Control "' + controlId + '": '+ dirty + ' - ', control);
+        // console.warn('  isDirty Control "' + controlId + '": '+ dirty + ' - ', control);
         return dirty;
     }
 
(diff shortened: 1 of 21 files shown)


src/client/app/shared/dynamic-form-fields/abstract-gnss-controls.ts
```diff
@@ -37,7 +37,7 @@ export abstract class AbstractGnssControls {
         } else {
             dirty = control.dirty;
         }
-        console.warn('  isDirty Control "' + controlId + '": '+ dirty + ' - ', control);
+        // console.warn('  isDirty Control "' + controlId + '": '+ dirty + ' - ', control);
```
lbodor: Remove this and all comment code.

oehm-smith: I'll want this when working on the validation so would prefer keep.  Also want the comments left in DateTime until I've fixed the remaining issues there.

lbodor:
:+1:

PR #366, opened 2017-05-18
hide UI components until teh data is loaded

heyagosper:


src/client/app/site-log/site-log.component.html
```diff
@@ -1,4 +1,7 @@
-<div class="container centered clear-pad">
+<div class="container centered clear-pad overlay" *ngIf="isLoading">
+    <div class="loading-img"></div>
+</div>
+<div class="container centered clear-pad" [hidden]="isLoading">
```
oehm-smith: I checked out the site and I found it interesting that despite saying `[hidden]`, the groups were still visible.  It turns out that this is not needed at all.  The image over the top along with the css overlay class seems to be enough to stop interaction with the page.

You can remove this line.

frankfu03:
There used to have a loading image running on the SiteLog page when loading is going on.

src/client/app/site-log/site-log.component.html
heyagosper: cool thanks

PR #370, opened 2017-05-19
add space to toolbar and remove padding from status

heyagosper:


oehm-smith:
:+1: :x:

frankfu03:
I think we should keep the toolbar height to a minimum.

PR #372, opened 2017-05-19
Fix new dates for Items to be the value at UTC.

oehm-smith:
HOLD OFF APPROVING - I've found a bug related to UTC time in the picker.

**Sorry, continue to do this PR.**  The problem is unrelated - it is when you have a new item, the control value doesn't yet exist so the DateTime code for setting the date hours, minutes, seconds fields comes from a new Date(), which of course is not the UTC value.

**I have fixed that bug now** as part of this PR.

### DIFF
--- a/src/client/app/humidity-sensor/humidity-sensor-view-model.spec.ts
+++ b/src/client/app/humidity-sensor/humidity-sensor-view-model.spec.ts
@@ -20,7 +20,7 @@ export function main() {
       expect(humiditySensorsViewModel.notes).toEqual('');
       expect(humiditySensorsViewModel.serialNumber).toEqual('');
       // The defaults for calibration and start date is now() - drop time when test for this
-      let nowPart: string = MiscUtils.getPresentDateTime().replace(/T.*/,'');
+      let nowPart: string = MiscUtils.getUTCDateTime().replace(/T.*/,'');
       expect(humiditySensorsViewModel.calibrationDate).toBeDefined();
       expect(humiditySensorsViewModel.calibrationDate).not.toEqual('');
       expect(humiditySensorsViewModel.calibrationDate).toContain(nowPart);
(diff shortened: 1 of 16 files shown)


oehm-smith:
Check it now team.

src/client/app/humidity-sensor/humidity-sensor-view-model.spec.ts
```diff
@@ -20,7 +20,7 @@ export function main() {
       expect(humiditySensorsViewModel.notes).toEqual('');
       expect(humiditySensorsViewModel.serialNumber).toEqual('');
       // The defaults for calibration and start date is now() - drop time when test for this
-      let nowPart: string = MiscUtils.getPresentDateTime().replace(/T.*/,'');
+      let nowPart: string = MiscUtils.getUTCDateTime().replace(/T.*/,'');
```
lbodor: There is no 'T' in the string returned by `getUTCDateTime()`, is there?

src/client/app/water-vapor-sensor/water-vapor-sensor-view-model.spec.ts
```diff
@@ -16,7 +16,7 @@ export function main() {
       expect(waterVaporSensorsViewModel.notes).toEqual('');
       expect(waterVaporSensorsViewModel.serialNumber).toEqual('');
       // The defaults for calibration and start date is now() - drop time when test for this
-      let nowPart: string = MiscUtils.getPresentDateTime().replace(/T.*/,'');
+      let nowPart: string = MiscUtils.getUTCDateTime().replace(/T.*/,'');
```
lbodor: There is no 'T' to replace, is there?

src/client/app/humidity-sensor/humidity-sensor-view-model.spec.ts
oehm-smith: This in a test.  Tests run fine.  The `replace` code is redundant so will remove.

frankfu03:
It looks all good now. Thanks. :+1:

PR #373, opened 2017-05-19
remove redundant dort direction support

heyagosper:


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -163,16 +162,10 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
...
-            this.itemProperties.push(item);
-            this.itemOriginalProperties.push(origItem);
-            indexAddedAt = this.itemProperties.length - 1;
-        } else {
-            this.itemProperties.splice(0, 0, item);
-            this.itemOriginalProperties.splice(0, 0, origItem);
-            indexAddedAt = 0;
-        }
+        let indexAddedAt: number = 0;
+        this.itemProperties.splice(0, 0, item);
```
oehm-smith: If you move the `indexAddedAt` then you might as well use it in the splice - first arg I believe.

PR #374, opened 2017-05-19
Remove unnecessary parsing to number

lbodor:
As defined by GeodesyML JSON schema, all relevant source values are already
numbers. Find relevant properties with `find src -name "*.ts" | xargs grep \'number\'`.


lbodor:
Superceded by #375.

PR #375, opened 2017-05-20
Redundant code

lbodor:


lbodor:
Note: Since submitting commit https://github.com/GeoscienceAustralia/GNSS-Site-Manager/pull/375/commits/41da8555d38d5a75fa2de55f3c6b5a3becb7f530, I've come to release that the date-picker, at run-time, stores values of type `Date` into our properties of type `string` and that therefore the removed code did in fact perform an important correction. I will let the commit stand, however, but close this and some subsequent pull requests  (#377 and #378), until I

1. Change the types of relevant properties in our view model to `Date`, so that no type change occurs at run-time. We shouldn't put up with properties changing type at run-time.
2. Re-implement date conversion functionality

PR #380, opened 2017-05-21
Map factory

lbodor:
Map-factory does everything JsonPointerService does and has support
for arrays as well, so we will be able to delete json-view-model.service.ts.


oehm-smith:
Bet you a billion that tests fail

oehm-smith:
Ok I was wrong.  The tests all pass.  But only 9 run!  The `test-config.js` and `karma-conf` need to be updated.

`22 05 2017 15:25:21.617:WARN [web-server]: 404: /base/node_modules/map-factory/dist/lib/map-factory.js`

PR #386, opened 2017-05-22
DateTime - new date defaults to UTC time

oehm-smith:
One problem with this is that the default date is at the time the dateTime component is created.


lbodor:
:+1:

PR #388, opened 2017-05-23
Geod 407 validation

oehm-smith:
* Improved Validation messages
* Fixed number custom component to take a max value and validate against min and max
* Fixed validation applied to item components

* The use of Validators.required causes a "was false now true" error, however this seems more of a warning - we must have fixed enough to prevent this being a fatal error.  I couldn't find a way to prevent this 'warning'.
* Not done was Validation for DateTime since ControlValueAccessor validation won't work for this unless it is modified.  HOWEVER I made some improvements to the DateTime around validation


oehm-smith:
Fixed validation 'bleed' issues and linting.

oehm-smith:
Ready for final review now

lbodor:
:) Good job!

PR #389, opened 2017-05-23
New item dirty flags

frankfu03:


lbodor:
:+1:

src/client/app/shared/abstract-groups-items/abstract-group.component.spec.ts
```diff
@@ -40,7 +41,7 @@ export function main() {
         beforeEach(() => {
             abstractGroupImpl = new AbstractGroupImpl(null);
             avmi1 = new AbstractViewModelImpl('4');
-            avmi1.setDateDeleted();
+            avmi1.setDateDeleted(MiscUtils.getUTCDateTime());
```
oehm-smith: This is a fairly significant change.  Why was it needed?

oehm-smith: Nope I see.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -204,26 +204,38 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
      * Remove an item.  Originally it was removed from the list.  However we now want to track deletes so
      * keep it and mark as deleted using change tracking.
      */
-    public removeItem(itemIndex: number, reason: string) {
-        // Be aware that the default order is one way (low to high start date), but what is displayed is the opposite
-        // (high to low start date).  This call is coming from the UI (the display order) and the default for
-        // getItemsCollection() is the reverse order so this works out ok
-        this.setDeletedReason(itemIndex, reason);
-        this.setDeleted(itemIndex);
+    public removeItem(index: number, reason: string) {
```
oehm-smith: I don't understand why this change would be made (I understand the code need to be modified).  I had organised the code into two methods so `removeItem()` reads - first `setDeletedReason`, then `set it as Deleted`.  Both of these contained the other code that has been added to update the `formGroup`.  Such as - `setDeletedReason()` says "find the item in the collection and update it with the reason; then update the equivalent form control".  The change makes it harder to understand the code.

src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
```diff
@@ -52,11 +52,19 @@ export class DatetimeInputComponent extends AbstractGnssControls implements OnIn
...
    /**
-    * Validate datetime value changed externally by manual typing.
-    #
-    * Note: it won't be invalid if it is selected by the calendar dialog
+    * Validate datetime value changed externally by manual typing or by JavaScript methods.
+    *
+    * Note: datetime values selected by the calendar dialog will be valid and dirty!
     */
     ngDoCheck(): void {
+        if(this.datetime !== this.formControl.value && this.formControl.value) {
```
oehm-smith: Should the guard `this.formControl.value` go first?

lbodor: Yes.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
lbodor: @oehm-smith, reading you old code I see what you mean. Moreover, the change is not strictly part of the bug fix. I'm happy for this to go back to how it was. I have two improvements to offer:

1) Rename `removeItem` to `deleteItem` -- there are almost synonymous, we are certainly not relying on any difference in meaning between them, so why use them both. It's already understood that delete is a logical delete, not an actual purging of data.

2) Like the old code factors out the if-guard `if (formGroup.controls['field'])`, we can also factor out the indexing of the `groupArrayForm`. Do it once and pass the item around, instead of the index.

lbodor: In the big scheme of things, however, we should probably just keep working on fixing bugs :)

oehm-smith: Yes, though this change was actively made and it would have required not much more effort to keep things aligned with the good style that had been setup.

lbodor: On the upside, it will let us weed out those setters that return anonymous objects of what was changed. As in:
```
setX(x: X): any {
    this.x = x;
    return { "x": x }
}
```

PR #390, opened 2017-05-23
Fix tests

lbodor:


src/client/app/shared/json-data-view-model/view-model/abstract-view-model.ts
```diff
@@ -95,7 +96,7 @@ export abstract class AbstractViewModel {
         this.dateDeleted = '';
         this.dateInserted = '';
         this.deletedReason = '';
-        this.startDate = '';
+        this.startDate = MiscUtils.getUTCDateTime();
```
oehm-smith: Is this change unrelated to the issue?

oehm-smith:
:+1:  though if that change is unrelated perhaps should go on a different PR.

src/client/app/shared/json-data-view-model/view-model/abstract-view-model.ts
lbodor: The tests fail without this change. It undoes a change Frank made yesterday. The tests don't run properly on the master, so we didn't know that Frank's change was braking them. We will ask Frank why he made that change, if that is the right thing to do, then we can also change the tests.

PR #392, opened 2017-05-24
Stop initialising dates in view models

lbodor:


src/client/app/humidity-sensor/humidity-sensor-view-model.ts
```diff
@@ -16,7 +15,7 @@ export class HumiditySensorViewModel extends AbstractViewModel {
      */
     constructor(blank: boolean = false) {
         super();
-        this.calibrationDate = blank ? '' : MiscUtils.getUTCDateTime();
+        this.calibrationDate = '';
```
oehm-smith: You should be able to get rid of `blank` throughout as well.  Since that was for the OrigSiteLog that is no longer being used.

src/client/app/pressure-sensor/pressure-sensor-view-model.ts
```diff
@@ -16,7 +15,7 @@ export class PressureSensorViewModel extends AbstractViewModel {
      */
     constructor(blank: boolean = false) {
```
oehm-smith: This is a `blank` that can be removed (confirm first!)

src/client/app/pressure-sensor/pressure-sensor-view-model.ts
```diff
@@ -16,7 +15,7 @@ export class PressureSensorViewModel extends AbstractViewModel {
...
         super();
-        this.calibrationDate = MiscUtils.getUTCDateTime();
+        this.calibrationDate = '';
```
oehm-smith: Can you remind me again where this is being set now?  StartDate is common and can be set when you `new()` but `calibrationDate` is unique to sensors.

lbodor: I won't be set. We will have to implement something new.

src/client/app/humidity-sensor/humidity-sensor-view-model.ts
lbodor: OK, thanks.

PR #394, opened 2017-05-24
GEOD-466 - Save should clear dirty aspects

oehm-smith:
* ItemComponents now stop appearing dirty and 'cancel' changed
to 'delete' after creating a new one and saving
* Changed Observed item from a single boolean representing dirty
condition of the form, to a state object that represents that
plus the save state of the application


lbodor:
:+1:

PR #395, opened 2017-05-24
Format outgoing dates

lbodor:


src/client/app/shared/json-data-view-model/data-view-translator.ts
```diff
@@ -66,8 +67,8 @@ export class DataViewTranslatorService {
             let targetPath = DataViewTranslatorService.toDotNotation(fieldMap.targetField.pointer);
 
             mapper.map(sourcePath).to(targetPath, (source: any) => {
-                if (source instanceof Date) {
-                    return MiscUtils.formatDateToDatetimeString(source as Date);
+                if (fieldMap.sourceField.type === 'date' && source !== '') {
+                    return moment.utc(source).format();
```
oehm-smith: I'd be explicit about the date format in case they change the default.

lbodor: OK, I'll make the change.

lbodor: Done.

PR #396, opened 2017-05-24
Stop traversing json geodesyml to remove nulls

lbodor:
Arrays like `[null]` are, since the switch to map-factory, no longer
generated.


oehm-smith:
:+1:

PR #399, opened 2017-05-24
GEOD-471 - Sensible Validation values

oehm-smith:
* Note that dates Validation must be handled directly in DateTimeComponent


lbodor:
:+1:

PR #401, opened 2017-05-25
Update invalid flag

frankfu03:


src/client/app/shared/status-info/status-info.component.ts
```diff
@@ -73,6 +73,7 @@ export class StatusInfoComponent implements OnInit {
             if (applicationState.applicationFormModified) {
                 this.isFormModified = true;
             }
+            this.isFormInvalid = applicationState.applicationFormInvalid;
```
oehm-smith: Be careful - it might be null and give results you don't expect.  Hopefully that sets the boolean to false.  Though this is why I put a guard around it.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -232,6 +232,7 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
             }
         }
 
+        (<FormGroup>this.groupArrayForm.at(itemIndex)).reset();
```
oehm-smith: Is there a `markAsValid()`?  `reset()` will set all values to default.  I guess this is in the `cancel` so that is not a problem.

oehm-smith:
Apart from you considering my comments, :+1:

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
frankfu03: Unfortunately there are no markAsValid() and markAsInvalid methods in Reactive Form. Yes, reset() is for cancelling new items, but it may cause problem when it reset required fields. SO far it works fine.

oehm-smith: Ok.  We'll go with this and keep this in mind.

src/client/app/shared/status-info/status-info.component.ts
frankfu03: Confirmed that "null" is the same as "false" in Boolean. The isFormInvalid flag need to track the changes between true and false. 

Actually, using a guard for isFormModified is incorrect. For example, once form is modified and isFormModified is set to true, it won't be able to set to false again when form is pristine. I can show the bug related to this guard.

oehm-smith: Good find; thanks Frank.

PR #406, opened 2017-05-25
Enable Prod mode to supress an error caused by empty values for requi…

frankfu03:


oehm-smith:
Oh I didn't know this is what we were going to do.  I vote no to this.  Fine for production, which the seed does autmoatically:

`src/client/app/main.ts:14:if (String('<%= BUILD_TYPE %>') === 'prod') { enableProdMode(); }`

But leave as an error in dev.  Since it doesn't seem to be causing any problems and it will remind us about it.  And hopefully we will see it disappear after some upgrade.

:-1:

lbodor:
Vetoed.

PR #407, opened 2017-05-25
GEOD-478 - Stop flash when not logged in and invalid

oehm-smith:


lbodor:
We can merge this pending demonstration that it does not cause any change detection errors.

oehm-smith:
I cannot reproduce the problem.  This is so bizarre.  I even went back to the commit before I started work on this.  Dev seems to be working fine now as well.  I can't explain.

PR #418, opened 2017-05-29
GEOD-431 - Add unsubscribers to prevent memory bloat

oehm-smith:
Based on https://stackoverflow.com/questions/38008334/angular-rxjs-when-should-i-unsubscribe-from-subscription, https://medium.com/@benlesh/rxjs-dont-unsubscribe-6753ed4fda87


lbodor:
Great :+1:

PR #419, opened 2017-05-29
set defaults and do not override nulls with blank string

heyagosper:


src/client/app/shared/dynamic-form-fields/text-input.component.ts
```diff
@@ -17,9 +17,6 @@ export class TextInputComponent extends AbstractGnssControls implements ControlV
     @Input() index: string = '0';
     @Input() name: string = '';
     @Input() label: string = '';
-    // controlName & form needed for validation
```
oehm-smith: Be sure to check that Validation still works after making these changes.

src/client/app/shared/json-data-view-model/data-view-translator.ts
```diff
@@ -67,12 +67,21 @@ export class DataViewTranslatorService {
             let targetPath = DataViewTranslatorService.toDotNotation(fieldMap.targetField.pointer) + '?';
 
             mapper.map(sourcePath).to(targetPath, (source: any) => {
-                if (source === undefined) {
```
oehm-smith: All this should be rewritable to:

```
if (! source) {
                    return null;
                }
                if (typeof source === 'string' && fieldMap.sourceField.type === 'date') {
                    return MiscUtils.formatUTCDateTime(source as string);
                }
                return source;
```

src/client/app/shared/dynamic-form-fields/text-input.component.ts
heyagosper: validation still works. :-)

src/client/app/shared/json-data-view-model/data-view-translator.ts
heyagosper: i don't want to do that. I started trying to make this code more elegant but I found that I was stepping on too many edge cases. Hence my comment about refactoring this once we do the task where we look at the model mapping code

oehm-smith:
You didn't like my suggested change?

Could we not `push -f` between reviews as that prevents one from seeing the changes made.

PR #423, opened 2017-05-30
gitGeod 431 cleanup poor memory use

oehm-smith:
* Minor change of something that bugged me in regard to memory allocation
* I will be putting up some documentation on memory profiling shortly (on eGeodesy wiki) - its at https://github.com/GeoscienceAustralia/egeodesy/wiki/Devs-~-Memory-profiling

### DIFF
--- a/src/client/app/shared/abstract-groups-items/abstract-group.component.ts
+++ b/src/client/app/shared/abstract-groups-items/abstract-group.component.ts
@@ -179,9 +179,9 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
     }
 
     setupChildItems() {
-        for (let viewModel of this.getItemsCollection()) {
+        this.getItemsCollection().forEach(() => {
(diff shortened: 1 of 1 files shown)


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -179,7 +179,8 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
     }
 
     setupChildItems() {
-        for (let viewModel of this.getItemsCollection()) {
+        let i: number = this.getItemsCollection().length;
+        for (; i >0; i--) {
             this.addChildItemToForm();
         }
     }
```
lbodor: `for (let i = 0; i < this.getItemsCollection().length; i++)`

lbodor: Or event better use lodash times.

PR #426, opened 2017-05-31
GEOD-487 Add aspiration field to temp sensor

oehm-smith:


lbodor:
Aspiration is still missing as a property from the view model class.

oehm-smith:
Thanks don't know how I missed that.

PR #429, opened 2017-05-31
Fix deleting and creating of responsible parties

lbodor:
The fix applies to site owner, metadata custodian, and data center.

### DIFF
--- a/src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
+++ b/src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
@@ -58,9 +58,9 @@ export class SiteLogDataModel {
             this.formInformation = this.getSuppliedValueOrEmptyValue(siteLogModel.formInformation, false);
             this.siteIdentification = this.getSuppliedValueOrEmptyValue(siteLogModel.siteIdentification, false);
             this.siteLocation = this.getSuppliedValueOrEmptyValue(siteLogModel.siteLocation, false);
-            this.siteOwner = this.getSuppliedValueOrEmptyValue(siteLogModel.siteOwner, false);
-            this.siteMetadataCustodian = this.getSuppliedValueOrEmptyValue(siteLogModel.siteMetadataCustodian, false);
-            this.siteDataSource = this.getSuppliedValueOrEmptyValue(siteLogModel.siteDataSource, false);
+            this.siteOwner = siteLogModel.siteOwner;
+            this.siteMetadataCustodian = siteLogModel.siteMetadataCustodian;
+            this.siteDataSource = siteLogModel.siteDataSource;
             this.moreInformation = this.getSuppliedValueOrEmptyValue(siteLogModel.moreInformation, false);
             this.dataStreams = this.getSuppliedValueOrEmptyValue(siteLogModel.dataStreams, false);
         }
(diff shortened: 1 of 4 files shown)


src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
```diff
@@ -58,9 +58,9 @@ export class SiteLogDataModel {
             this.formInformation = this.getSuppliedValueOrEmptyValue(siteLogModel.formInformation, false);
             this.siteIdentification = this.getSuppliedValueOrEmptyValue(siteLogModel.siteIdentification, false);
             this.siteLocation = this.getSuppliedValueOrEmptyValue(siteLogModel.siteLocation, false);
-            this.siteOwner = this.getSuppliedValueOrEmptyValue(siteLogModel.siteOwner, false);
-            this.siteMetadataCustodian = this.getSuppliedValueOrEmptyValue(siteLogModel.siteMetadataCustodian, false);
-            this.siteDataSource = this.getSuppliedValueOrEmptyValue(siteLogModel.siteDataSource, false);
+            this.siteOwner = siteLogModel.siteOwner;
```
oehm-smith: So it is null if it doesn't exist?  Perhaps a method similar to the other would follow the same style and read better?

lbodor: What would the method do?

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -87,17 +95,30 @@ export class JsonViewModelService {
...
         siteLogDataModel.siteContacts = this.viewToDataModel(viewModel.siteContacts);
 
-        DataViewTranslatorService.translate(viewModel.siteDataSource[0], siteLogDataModel.siteDataSource,
-            new ResponsiblePartyViewModel().getObjectMap().inverse());
+        if (viewModel.siteDataSource && viewModel.siteDataSource.length !== 0) {
+            if (!siteLogDataModel.siteDataSource) {
+                siteLogDataModel.siteDataSource = {};
+            }
+            DataViewTranslatorService.translate(viewModel.siteDataSource[0], siteLogDataModel.siteDataSource,
+                new ResponsiblePartyViewModel().getObjectMap().inverse());
```
oehm-smith: So the `ObjectMap` represents the DataToView normally but inversed to ViewToData.  Perhaps a better name to signify this is warranted?

lbodor: I see what you mean, if it were called DataToViewMap, for example, then taking the inverse would make it obvious that what you are getting as a ViewToDataMap. I think that we should make that change a little later, perhaps when we attempt to pull out all the mappings from the view models into one big map.

src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
oehm-smith: `this.getSuppliedValueOrUndefined(..)`

src/client/app/shared/json-data-view-model/json-view-model.service.ts
oehm-smith: :+1:

src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
lbodor: That's what it would be called, not what it would do. Could we see the body? :)

oehm-smith:
:+1:

src/client/app/shared/json-data-view-model/data-model/site-log-data-model.ts
oehm-smith:
```
private getSuppliedValueOrUndefined(value : any) : any {
        if (value !== null && value !== undefined) {
           return value;
        } else {
           return undefined;
        }
    }
```

PR #434, opened 2017-06-01
Remove ability to hide deleted items

lbodor:
Something here was interfering with our recent initialisation of
`itemProperties`, so that we couldn't create items in empty collections.


oehm-smith:
:+1:

PR #435, opened 2017-06-01
GEOD-497 - Add Signal Obstruction

oehm-smith:
* Confirmed can load, edit save
* Lazar is working on separate issue to do with
creating new items


lbodor:
:+1:

PR #437, opened 2017-06-01
Simplify

lbodor:


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -114,7 +114,7 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
     }
 
     setItemsCollection(itemProperties: T[]) {
-        if (itemProperties && itemProperties.length > 0) {
```
oehm-smith: Test this as at one stage this guard was necessary.

PR #441, opened 2017-06-05
Fix GEOD-502, delete site owner bug

lbodor:


oehm-smith:
:+1:  except for failed build

PR #443, opened 2017-06-05
Tidy up

lbodor:


src/client/app/shared/global/user-auth.service.ts
```diff
@@ -23,9 +23,9 @@ export class UserRegistration {
  */
 @Injectable()
 export class UserAuthService {
-    public userLoadedEvent: EventEmitter<User> = new EventEmitter<User>();
+    public userLoadedEvent = new EventEmitter<User | null>();
```
oehm-smith: Perhaps you need to make it:

```public userLoadedEvent: EventEmitter<any> = new ...`

File: src/client/app/shared/global/user-auth.service.ts [hunk 2]
```diff
@@ -118,8 +113,9 @@ export class UserAuthService {
     }
 
     private addEventHandlers() {
-        this.userManager.events.addUserUnloaded((e) => {
+        this.userManager.events.addUserUnloaded(() => {
             this.currentUser = null;
+            this.userLoadedEvent.emit(null);
```
oehm-smith: Was it meant to be `emit(e)`?

src/client/app/shared/global/user-auth.service.ts
lbodor: I think that the compiler can infer that.

src/client/app/shared/global/user-auth.service.ts
lbodor: I don't think that there is an `e`.

PR #447, opened 2017-06-06
allow readonly attribute and set for 4 char id

heyagosper:
Note, I had thought that 'readonly=""' was invalid but it is valid html5 to use an empty string.
I still didn't work out why I can't set it to the value "readonly" in the html element but this works fine and is conformant so there's no real point figuring it out other than as a learning exercise.

### DIFF
--- a/src/client/app/shared/dynamic-form-fields/text-input.component.html
+++ b/src/client/app/shared/dynamic-form-fields/text-input.component.html
@@ -6,6 +6,7 @@
     <div class="text-input" class="col-md-5 col-sm-6 col-xs-6 col-xxs-12">
          <input (keyup.enter)="ignoreEvent($event);"
              type="text"
+             readonly="{{readonly}}"
              [formControlName]="controlName"
              class="form-control"
              />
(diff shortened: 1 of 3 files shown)


lbodor:
How will we make the four character id editable for new sites?

src/client/app/shared/dynamic-form-fields/text-input.component.html
```diff
@@ -6,6 +6,7 @@
     <div class="text-input" class="col-md-5 col-sm-6 col-xs-6 col-xxs-12">
          <input (keyup.enter)="ignoreEvent($event);"
              type="text"
+             readonly="{{readonly}}"
```
oehm-smith: I see where you are going though officially this should be `[readonly]="readonly"`

src/client/app/site-log/site-identification.component.html
```diff
@@ -20,7 +20,7 @@
         </div>
         <div class="margin-bottom15" *ngIf=" siteIdentification != null ">
             <div class="form-group">
-                <text-input formControlName="fourCharacterID" controlName="fourCharacterID"
+                <text-input readonly="readonly" formControlName="fourCharacterID" controlName="fourCharacterID"
```
oehm-smith: Like Lazar asked, you need a way to make this "" so that the field can be editable.  Thus have a variable and use `[readonly]="readonly"` again.

PR #451, opened 2017-06-08
Add linting rule for unused imports

oehm-smith:


oehm-smith:
The build will fail as the new tslint rule will run.  I will fix those problems next.

PR #460, opened 2017-06-09
Geod 506 blank loc pos on load no required

oehm-smith:
This change is to set validators on the Geodetic and Cartesian Position Groups in SiteLocation.  So that the associated fields are either all required or none are (once a value goes in, all fields must be filled).  It also allows for additional validators to be applied.

Apologies that appears to be lots of changes in `data-view-translator.spec.ts` but the only changes were in:

       describe('Point type translate', () => { ... }

This was caused by the merge / rebase.

### DIFF
--- a/src/client/app/site-log/site-location.component.ts
+++ b/src/client/app/site-log/site-location.component.ts
@@ -1,11 +1,12 @@
-import { Component, OnInit, Input, ChangeDetectorRef } from '@angular/core';
-import { FormBuilder, FormGroup, Validators, AbstractControl } from '@angular/forms';
+import { Component, OnInit, Input } from '@angular/core';
+import { FormBuilder, FormGroup, Validators, AbstractControl, ValidatorFn } from '@angular/forms';
 import { MiscUtils } from '../shared/global/misc-utils';
 import { DialogService } from '../shared/index';
 import { UserAuthService } from '../shared/global/user-auth.service';
 import { SiteLogService, ApplicationState, ApplicationSaveState } from '../shared/site-log/site-log.service';
 import { AbstractBaseComponent } from '../shared/abstract-groups-items/abstract-base.component';
 import { SiteLocationViewModel } from './site-location-view-model';
+import * as _ from 'lodash';
 
 /**
  * This class represents the SiteLocation sub-component under the SiteInformation Component.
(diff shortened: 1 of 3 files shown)


src/client/app/site-log/site-location.component.ts
```diff
@@ -121,25 +121,25 @@ export class SiteLocationComponent implements OnInit {
...
-    }
-
-    private addRequiredValidator(itemControl: AbstractControl, controlName: string) {
-        itemControl.setValidators(Validators.required);
+    /**
+     * This method is to determine whether a required validator is present.
+     * @param validatorFn - the function as created by the Angular framework
+     */
+    private hasRequiredValidator(validatorFn: ValidatorFn | null): boolean {
+        return validatorFn ? validatorFn === Validators.required : false ;
```
lbodor: Let's not use ternary conditionals to generate a booleans, instead write `validatorFn && validatorFn === Validators.required`.

lbodor: In this case, would `validatorFn === Validators.required` work just as well?

src/client/app/site-log/site-location.component.ts
```diff
@@ -121,25 +121,25 @@ export class SiteLocationComponent implements OnInit {
                 throw new Error(`Control for group item: ${key} doesnt exist in positionGroup: ${positionGroup}`);
             }
             if (someFieldHasValue) {
-                if (!itemControl.validator) {
-                    this.addRequiredValidator(itemControl, key);
+                if (!this.hasRequiredValidator(itemControl.validator)) {
+                    itemControl.setValidators(Validators.required);
```
lbodor: What if there are multiple validators? Would this blow away all existing validators? If so, perhaps we should be adding and removing required validators, rather than setting.

lbodor:
Rewrite your commits please to remove conflict markers.

src/client/app/site-log/site-location.component.ts
oehm-smith: No it might be null.

lbodor: I don't think that's a problem, `null === Validators.required` is `false`. I would recommend removing `hasRequiredValidator` function altogether, if all it ends up doing is `validatorFn === Validators.required`.

oehm-smith:
All good to review again.

lbodor:
Could you please squash your commits. There is no need for commits to add conflict markers and then remove them. Commits should be logical where possible, rather than historical or in this case accidental. I know that that will make the PR comments harder to follow, but it's worth it. Think of git bisect, for example, encountering commits that don't compile because of committed conflict markers.

PR #463, opened 2017-06-09
Remove unneeded ternary operators

lbodor:
Generate boolean values using boolean expressions, not ternary
operators.


frankfu03:
:+1:

PR #466, opened 2017-06-12
Update user profile

frankfu03:


src/client/app/shared/toolbar/toolbar.component.ts
```diff
@@ -122,6 +126,10 @@ export class ToolbarComponent implements OnInit, OnDestroy {
         this.userAuthService.changePassword();
     }
 
+    private getValue(value: any): string {
+        return value ? value : '';
```
lbodor: The following idiom is common in Javascript: `return value || ''`. It's shorter and does the same job.

lbodor:
Good choice removing the helper function :+1:

src/client/app/shared/toolbar/toolbar.component.ts
frankfu03: Yes, good suggestion. Relevant changes have been made. Thanks.

PR #472, opened 2017-06-13
update registration form components including validation

heyagosper:


src/client/app/user-registration/user-registration.component.ts
```diff
@@ -28,7 +28,7 @@ export class UserRegistrationComponent implements OnDestroy {
         private location: Location,
         private formBuilder: FormBuilder,
         private userService: UserAuthService,
-        private dialogs: DialogService,
+        private dialogs: DialogService
```
lbodor: I think that in most languages trailing commas are well regarded.

oehm-smith:
:+1:  :package:

PR #473, opened 2017-06-13
Prefetch data before rendering siteLog form

frankfu03:


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -18,14 +18,7 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
...
     @Input() parentForm: FormGroup;
-
-    @Input()
-    set siteLogModel(siteLogModel: any) {
-       if (siteLogModel) {
-           this.setItemsCollection(this.getFormData(siteLogModel));
-           this.setupChildItems();
-       }
-    }
+    @Input('siteLogModel') siteLogModel: T;
```
lbodor: I don't think `T` is right type for `siteLogModel`. `T` is constrained as `T extends AbstractViewModel` (so it's different for each subclass of `AbstractGroupComponent`), but siteLogModel is always the top-level `SiteLogViewModel`.

src/client/app/shared/site-log/prefetch-site-log.service.ts
```diff
@@ -0,0 +1,44 @@
...
+            return Promise.resolve(new SiteLogViewModel());
+        }
+
+        return this.http.get(this.constantsService.getWebServiceURL()
+                 + '/siteLogs/search/findByFourCharacterId?id=' + fourCharacterId + '&format=geodesyml')
+            .toPromise().then((result: any) => {
+                var geodesyMl: any = result.text();
+                let json: string = this.jsonixService.geodesyMLToJson(geodesyMl);
+                let siteLogViewModel: SiteLogViewModel = this.jsonViewModelService.dataModelToViewModel(json);
+                return siteLogViewModel;
```
lbodor: Could we call `SiteLogService` to fetch the site log?

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
frankfu03: Yeah. I will correct it soon.

src/client/app/shared/site-log/prefetch-site-log.service.ts
frankfu03: I tried, but failed. I think that it is related to the return type which requires a Promise, not a Observable. If needed, I can change getSiteLogByFourCharacterIdUsingGeodesyML() method in SiteLogService.

lbodor: You can convert an `Observable` to a `Promise`. See https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/topromise.md.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -18,14 +19,7 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
...
     @Input() parentForm: FormGroup;
-
-    @Input()
-    set siteLogModel(siteLogModel: any) {
-       if (siteLogModel) {
-           this.setItemsCollection(this.getFormData(siteLogModel));
-           this.setupChildItems();
-       }
-    }
+    @Input('siteLogModel') siteLogModel: SiteLogViewModel;
```
lbodor: Would `@Input() siteLogModel: SiteLogViewModel;` be sufficient?

src/client/app/shared/site-log/prefetch-site-log.service.ts
```diff
@@ -0,0 +1,35 @@
...
+            this.router.navigate([homeUrl]);
+            return null;
+        } else if (fourCharacterId === 'newSite') {
+            return Promise.resolve(new SiteLogViewModel());
+        }
+
+        return this.siteLogService.getSiteLogByFourCharacterId(fourCharacterId)
+            .then((result: any) => {
+                let siteLogViewModel: SiteLogViewModel = result;
+                return siteLogViewModel;
```
lbodor: If types are the reason for the local declaration, then you can try `return result as SiteLogViewModel;`.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
frankfu03: Yeah, it is the same in this case because that the variable name used in the component is the same as the one passed from directive. We may remove it though it has no harm.

src/client/app/shared/site-log/site-log.service.ts
```diff
@@ -51,6 +51,35 @@ export class SiteLogService implements OnDestroy {
...
+                            console.error('SiteLogService - Failed in fetching siteLog by FourCharacterId: ', error);
+                            reject(null);
+                        }
+                    );
+            } catch (error) {
+                console.error('SiteLogService - Error in fetching siteLog by FourCharacterId: ', error);
+                reject(null);
+            }
+        });
+    }
```
lbodor: Have you considered calling `toPromise()` on the observable instead of creating the promise manually?

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
lbodor: No harm.

lbodor:
:+1:  I'm really glad that pre-fetching works.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -168,6 +162,8 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel> extend
             this.parentForm.removeControl(itemsArrayName);
         }
         this.parentForm.addControl(itemsArrayName, this.groupArrayForm);
+        this.setItemsCollection(this.getFormData(this.siteLogModel));
```
oehm-smith: I don't know if your new approach avoids the problem I had that moved the setter call to the @Input.  The timing is out (maybe this is fixed) and even though the @Input is called first to set `siteLogModel`, it is initially null.  Thus why there was a guard above.  Kudos to you / your new method if this is no longer a problem.

src/client/app/shared/site-log/prefetch-site-log.service.ts
```diff
@@ -0,0 +1,35 @@
...
+                private dialogService: DialogService,
+                private siteLogService: SiteLogService) {
+    }
+
+    resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<SiteLogViewModel> {
+        let homeUrl: string = '/';
+        let fourCharacterId: string = route.params['id'];
+        if (!fourCharacterId) {
+            this.router.navigate([homeUrl]);
+            return null;
```
oehm-smith: You need to return a promise that returns null (I would think).

src/client/app/shared/site-log/site-log.service.ts
oehm-smith: Why create promises when we've been using Observables all along?

src/client/app/shared/site-log/prefetch-site-log.service.ts
frankfu03: Yeah, I simplied it to "return <SiteLogViewModel>result;" and it works fine.

frankfu03: I means "return \<SiteLogViewModel\>result";

src/client/app/shared/site-log/site-log.service.ts
frankfu03: Resolve is used to prefetch sitelog and it returns Promise. We may call toPromise() in a separate branch.

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
frankfu03: Because site log data has been prefetched, it seems that @Input is called before ngOnInit(). So we may set values to all form fields in setupForm() for items.

src/client/app/shared/site-log/prefetch-site-log.service.ts
frankfu03: It seems working fine at the moment. Not sure if I should change it to "return Promise.resolve(null);"?

src/client/app/shared/abstract-groups-items/abstract-group.component.ts
oehm-smith: `@Input` was always called before `ngOnInit()` but just that the first time its argument was null.  The prefetch mechanism seems to defer the creation of the component and thus the data is available.  Good work!

src/client/app/shared/site-log/prefetch-site-log.service.ts
oehm-smith: I wouldn't think you'd do that.  More like this:

    return new Promise((resolve, reject) => {
        let homeUrl: string = '/';
        let fourCharacterId: string = route.params['id'];
        if (!fourCharacterId) {
            this.router.navigate([homeUrl]);
            resolve(null);
        } else if (fourCharacterId === 'newSite') {
            return resolve(new SiteLogViewModel());
        }

src/client/app/shared/site-log/site-log.service.ts
frankfu03: I checked the API doc for Resolve. it supports Promise, Observable, and any Object. I tried using Observable but had problem in catching errors. So I may just leave it at the moment.

src/client/app/shared/site-log/prefetch-site-log.service.ts
frankfu03: "return null" is fine for Resolve as stated in API doc. Your code is another good approach.

src/client/app/shared/site-log/site-log.service.ts
oehm-smith: I'm updating the comment in prefetch below to use Observable instead.

src/client/app/shared/site-log/prefetch-site-log.service.ts
oehm-smith: Ok.  Though I think we should do it properly.  But up to you and Lazar.  I also think Observables should be used instead and I'll write a sample below.  Promises give nothing over Observables except 1 less line of code (Observables use `complete()`.  (ALthought it looks like I didn't use this in the placed I had written Observables.  The code should be updated for those.  I'll put in an Issue.). And if Observables are used then streaming becomes an option.

    return new Observable((observer: Subscriber<any>) => {
        let homeUrl: string = '/';
        let fourCharacterId: string = route.params['id'];
        if (!fourCharacterId) {
            this.router.navigate([homeUrl]);
            observer.next(null);
        } else if (fourCharacterId === 'newSite') {
            observer.next(new SiteLogViewModel());
        }
        observer.complete();
        // and if an error somewhere use
        // observer.error(new Error(the_error));
    })

lbodor: If the API says that the return type is `Observable<T> | Promise<T> | T`, then returning `new SiteLogViewModel()`, not wrapped in either a promise or an observable is proper. It is also the simplest approach. Observables can stream, but streaming is not need here.

I think that it's great that pre-fetching works and that it can simplify our code in other places. Why don't we merge this now, and try to use an observable end-to-end for the asynchronous case in a future pull request, when we've learned how to solve the problem Frank encountered. For the synchronous case, returning a plain `SiteLogViewModel` is OK according to the API.

PR #475, opened 2017-06-13
Geod 498 make 4character id field readonly

heyagosper:


src/client/app/shared/toolbar/toolbar.component.ts
```diff
@@ -70,8 +70,17 @@ export class ToolbarComponent implements OnInit, OnDestroy {
...
+     * save is always enabled for new site because everything is optional
+     * save is disabled if nothing has in the form has changed or the user is not allowed to edit this site
+     */
     public isSaveDisabled(): boolean {
-        return !(this.hasAuthorityToEditSite() && this.isFormDirty());
+        let isSaveDisabled = false;
+        if (this.siteId !== 'newSite') {
+            isSaveDisabled = !this.isFormDirty() || !this.hasAuthorityToEditSite();
+        }
+        return isSaveDisabled;
```
lbodor: I think that save should be disabled for new sites when the form is not dirty. Why save an empty form?

lbodor:
:+1:

lbodor:
:+1:

src/client/app/shared/dynamic-form-fields/text-input.component.html
```diff
@@ -6,6 +6,7 @@
     <div class="text-input" class="col-md-5 col-sm-6 col-xs-6 col-xxs-12">
          <input (keyup.enter)="ignoreEvent($event);"
              type="text"
+             readonly="{{readonly}}"
```
oehm-smith: Square brackets instead.  Yesterday when you had a similar thing for invalid(), the {{}} doesn't seem to refresh all the tiem.

PR #477, opened 2017-06-13
Move contact ObjectMaps into JsonViewModelService

lbodor:
Eventually all ObjectMaps will move out of data objects.

### DIFF
--- a/src/client/app/shared/json-data-view-model/data-view-translator.ts
+++ b/src/client/app/shared/json-data-view-model/data-view-translator.ts
@@ -18,33 +18,102 @@ export class TypedPointer {
  * mapping to translate one of view or data model to the other.
  */
 export class FieldMap {
-    constructor(public readonly sourceField: TypedPointer,
-                public readonly targetField: TypedPointer) {}
+    constructor(
+        public readonly sourceField: TypedPointer,
+        public readonly targetField: TypedPointer,
+        public readonly objectMap: ObjectMap = null,
+    ) {}
 
     public inverse(): FieldMap {
-        return new FieldMap(this.targetField, this.sourceField);
+        return new FieldMap(this.targetField, this.sourceField, this.objectMap ? this.objectMap.inverse() : null);
     }
 }
+
+type MapFn = (a: any) => any;
 
 export class ObjectMap {
 
     private fieldMaps = new Array<FieldMap>();
+
+    private sourcePreMap: MapFn = _.identity;
+    private targetPreMap: MapFn = _.identity;
+
+    private sourcePostMap: MapFn = _.identity;
+    private targetPostMap: MapFn = _.identity;
 
     public inverse(): ObjectMap {
         let inverse = new ObjectMap();
         _.forEach(this.fieldMaps, f => {
             inverse.add(f.inverse());
         });
+        inverse.sourcePreMap = this.targetPreMap;
+        inverse.targetPreMap = this.sourcePreMap;
+        inverse.sourcePostMap = this.targetPostMap;
+        inverse.targetPostMap = this.sourcePostMap;
         return inverse;
+    }
+
+    public addSourcePreMap(fn: MapFn): ObjectMap {
+        this.sourcePreMap = _.flow(this.sourcePreMap, fn);
+        return this;
+    }
+
+    public addTargetPreMap(fn: MapFn): ObjectMap {
+        this.targetPreMap = _.flow(this.targetPreMap, fn);
+        return this;
+    }
+
+    public addSourcePostMap(fn: MapFn): ObjectMap {
+        this.sourcePostMap = _.flow(this.sourcePostMap, fn);
+        return this;
+    }
+
+    public addTargetPostMap(fn: MapFn): ObjectMap {
+        this.targetPostMap = _.flow(this.targetPostMap, fn);
+        return this;
     }
 
     public add(fieldMap: FieldMap) {
         this.fieldMaps.push(fieldMap);
     }
 
+    public addFieldMap(source: string, target: string, objectMap: ObjectMap = null): ObjectMap {
+        this.add(new FieldMap(new TypedPointer(source, null), new TypedPointer(target, null), objectMap));
+        return this;
+    }
+
     // TODO: remove this getter once object translate is implemented in ObjectMap
     public getFieldMaps(): FieldMap[] {
         return this.fieldMaps;
+    }
+
+    public map(source: any): any {
+        source = this.sourcePreMap(source);
+        let mapped = source ? this.getObjectMapper().execute(source) : null;
+        return this.targetPostMap(mapped);
+    }
+
+    private getObjectMapper(): any {
+        let mapper = createMapper({ alwaysTransform: true, alwaysSet: true });
+        for (let fieldMap of this.fieldMaps) {
+            let sourcePath = fieldMap.sourceField.pointer;
+            let targetPath = fieldMap.targetField.pointer + '?';
+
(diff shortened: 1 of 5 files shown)


src/client/app/shared/json-data-view-model/data-view-translator.ts
```diff
@@ -18,34 +18,103 @@ export class TypedPointer {
  * mapping to translate one of view or data model to the other.
  */
 export class FieldMap {
```
oehm-smith: The changes here are quite complex.  Could you add spec tests to help describe the use.

src/client/app/shared/json-data-view-model/data-view-translator.ts
```diff
@@ -18,34 +18,103 @@ export class TypedPointer {
...
 
     public inverse(): FieldMap {
-        return new FieldMap(this.targetField, this.sourceField);
+        return new FieldMap(this.targetField, this.sourceField, this.objectMap ? this.objectMap.inverse() : null);
     }
 }
 
+type MapFn = (a: any) => any;
+
 export class ObjectMap {
```
oehm-smith: This is big enough to warrant its own file.

lbodor: I will eventually take over the entire file, removing everything from it that is not related to object maps. Then, I will rename the file.

src/client/app/shared/json-data-view-model/data-view-translator.ts
lbodor: Yes, good idea. I'd prefer to wait, just a little longer, until few more object maps are converted, to let the approach stabilise so to avoid documenting twice. The existing tests that you wrote are still there and still helpful.

oehm-smith: Your approach makes sense though you might leave people behind if not done soon.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -9,15 +9,71 @@ import { HumiditySensorViewModel } from '../../humidity-sensor/humidity-sensor-v
...
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.address.ciAddress.administrativeArea.characterString.gco:CharacterString', 'administrativeArea')
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.address.ciAddress.postalCode.characterString.gco:CharacterString', 'postalCode')
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.address.ciAddress.city.country.characterString.gco:CharacterString', 'country')
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.address.ciAddress.electronicMailAddress[0].characterString.gco:CharacterString', 'email')
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.phone.ciTelephone.voice[0].characterString.gco:CharacterString', 'phone')
+    .addFieldMap('ciResponsibleParty.contactInfo.ciContact.phone.ciTelephone.facsimile[0].characterString.gco:CharacterString', 'fax')
+;
+/* tslint:disable:max-line-length */
+
+function removeNullsFromArrays(obj: Object): void {
```
oehm-smith: Why not make these methods?

lbodor: As before, I feel it's just a little too soon to know where to put them. I'd prefer to delay making the decision where to put this function until it becomes obvious. This PR is really a sneak preview of changes to come. I could sit on it longer, but I need testers :)

What are the immediate negative consequences of leaving these functions the way there are?

oehm-smith: I don't think there is any problem with functions excepting you've written a class so generally methods are the norm, and you don't have a this context.  Of course the transpiler makes methods into functions regardless.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -9,15 +9,71 @@ import { HumiditySensorViewModel } from '../../humidity-sensor/humidity-sensor-v
...
 import { SiteLocationViewModel } from '../../site-log/site-location-view-model';
 import { RadioInterferenceViewModel } from '../../radio-interference/radio-interference-view-model';
 import { SignalObstructionViewModel } from '../../signal-obstruction/signal-obstruction-view-model';
 import { MultipathSourceViewModel } from '../../multipath-source/multipath-source-view-model';
+import * as _ from 'lodash';
+
+/* tslint:disable:max-line-length */
+let responsiblePartyMap = new ObjectMap()
+
+    .addSourcePreMap((source: any): any => {
```
oehm-smith: Isnt this just the _identity function, which is the default.

lbodor: No, it's not the identity. Identity returns its argument, this function sometimes returns its argument and sometimes `null`.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
lbodor: But, I haven't written a class yet, that's the issue. I'm unsure yet what that class will be, so I'm leaving the function at top level. I do know that this function doesn't belong to `ObjectMap`, but to a client of `ObjectMap.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -9,15 +9,71 @@ import { HumiditySensorViewModel } from '../../humidity-sensor/humidity-sensor-v
...
+}
+
+let siteLogMap = new ObjectMap()
+    .addFieldMap('siteOwner', 'siteOwner[0]', responsiblePartyMap)
+    .addFieldMap('siteContacts', 'siteContacts', responsiblePartyMap)
+    .addFieldMap('siteMetadataCustodian', 'siteMetadataCustodian[0]', responsiblePartyMap)
+    .addFieldMap('siteDataCenters', 'siteDataCenters', responsiblePartyMap)
+    .addFieldMap('siteDataSource', 'siteDataSource[0]', responsiblePartyMap)
+
+    .addTargetPostMap((target: any): any => {
```
oehm-smith: Should that be `any[]`

lbodor: No, it operators on the target of siteLogMap, which is the entire view model.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -100,39 +144,11 @@ export class JsonViewModelService {
...
-                siteLogDataModel.siteMetadataCustodian = {};
-            }
-            DataViewTranslatorService.translate(viewModel.siteMetadataCustodian[0], siteLogDataModel.siteMetadataCustodian,
-                new ResponsiblePartyViewModel().getObjectMap().inverse());
-        }
-
         siteLogDataModel.moreInformation = viewModel.moreInformation;
         siteLogDataModel.dataStreams = viewModel.dataStreams;
 
+        _.merge(siteLogDataModel, siteLogMap.inverse().map(viewModel));
```
oehm-smith: If you have this do you need the above lines?

lbodor: The new mapper handles only responsible parties. When we convert all view models using the new approach, then the above lines will not be needed. Notice that the old handling or responsible parties is gone from above.

oehm-smith:
:+1:

PR #479, opened 2017-06-14
use square-bracket syntax for readonly attribute

heyagosper:


src/client/app/shared/dynamic-form-fields/text-input.component.ts
```diff
@@ -37,6 +37,10 @@ export class TextInputComponent extends AbstractGnssControls implements ControlV
         this.propagateTouch = fn;
     }
 
+    public getReadonlyAttribute(): string {
+    	return this.readonly ? this.readonly : null;
+    }
```
lbodor: This function seems to effectively convert an empty string to `null`. What are the possible significant values of `readonly`? Could we initialise `readonly` to `null` instead, so we don't have to convert?

heyagosper: done

lbodor:
I've modified the linter to accept `!=` and `==` when comparing against `null`, because that takes care of `undefined` as well, via type coercion. See: #478. `null` and `undefined` coerce to each other. I don't know what's best, except that the developer should make the call in this case, not the linter.

PR #483, opened 2017-06-14
Use patchValue to avoid hidden form controls

lbodor:
Currently, all abstract view items have a property called `objectMap`,
which is not a data field. Using `setValue` forces us to add such properties
to reactive forms as hidden controls.


oehm-smith:
I would have thought we'd prefer ObjectMap is added as a field. `patchValue()` is much forgiving compared to `setValue()` and could allow problems, such as a form or data field to be missing.

lbodor:
We should probably switch back to `setValue` for the reason you mention. In the meantime, while we are trying to remove `objectMap` from components one at a time, the leniency of `patchValue` is helpful. It's not just object maps, responsible parties do not really have any date fields, so `setValue` hurts us there too. I think that patching is a better intermediate fix than adding fields or controls to where they don't belong. It's why `patchValue` exists in the reactive forms API.

PR #484, opened 2017-06-14
Refactor antenna object map

lbodor:


src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -40,6 +40,30 @@ let responsiblePartyMap = new ObjectMap()
 ;
 /* tslint:disable:max-line-length */
 
+let dateMap = new ObjectMap().addSourcePostMap((source: string): string => {
+    return source ? MiscUtils.formatUTCDateTime(source) : null;
+});
+
+let gnssAntennaMap = new ObjectMap()
+    .addFieldMap('dateDeleted.value[0]', 'dateDeleted', dateMap)
```
oehm-smith: Have you got something in mind as to how to abstract these 3 common fields?

src/client/app/shared/json-data-view-model/json-view-model.service.ts
```diff
@@ -69,6 +93,8 @@ let siteLogMap = new ObjectMap()
     .addFieldMap('siteDataCenters', 'siteDataCenters', responsiblePartyMap)
     .addFieldMap('siteDataSource', 'siteDataSource[0]', responsiblePartyMap)
 
+    .addFieldMap('gnssAntennas', 'gnssAntennas', gnssAntennaMap)
```
oehm-smith: Missing ;

src/client/app/shared/json-data-view-model/data-view-translator.ts
```diff
@@ -89,7 +89,14 @@ export class ObjectMap {
 
     public map(source: any): any {
         source = this.sourcePreMap(source);
-        let mapped = source ? this.getObjectMapper().execute(source) : null;
+        let mapped: any;
```
oehm-smith: Can we avoid 'any' with an abstract class?

src/client/app/shared/json-data-view-model/json-view-model.service.ts
lbodor: I don't think so. A semicolon at the end of that line would be a syntax error. Also, the linter runs automatically during the build process.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
lbodor: No, nothing concrete yet. One option would be to set up an object map with the common fields and make that the base of other object maps by taking a copy. Another option would be to implement composition of object maps, to effectively apply multiple object maps in sequence.

src/client/app/shared/json-data-view-model/data-view-translator.ts
lbodor: I'm not sure what you mean. What abstract class?  Object mappers are meant to map object for which we don't have types, like our GeodesyML DTOs, which are generated by Jsonix.

oehm-smith: What I mean is that you haven't typed this.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
oehm-smith: Something you'll do later? It would be easier to work this out early on.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
oehm-smith: Oh I see; the newline threw me - there is more following.

src/client/app/shared/json-data-view-model/data-view-translator.ts
lbodor: Object mappers are meant to work with objects of any type. Our GeodesyML DTOs  do not have types, and neither do reactive form values. If we are going to leave open the option of deleting our view model classes and mapping directly from the DTOs to the form controls (both are untyped), then I think we shouldn't jump the gun with type annotations.

Also, we don't impose any static requirements on the mapped values, so types would not help us enforce any requirements.

src/client/app/shared/json-data-view-model/json-view-model.service.ts
lbodor: Yes, soon. I outlined two options above. As we stand, there is no need for either yet, because there is no code duplication yet :) Also, in my opinion, it is better to put up for a while with some amount of code duplication, then to reach for a bad solution too soon -- the cost of being wrong is smaller.

PR #485, opened 2017-06-14
Use reactive form raw value

lbodor:
The raw value contains disabled fields, which we will need to
in order to save delete log items.


oehm-smith:
Right-o.  I never knew that. :+1:

oehm-smith:
I'll rebase and then merge.

PR #489, opened 2017-06-16
e2e, use host network mode

lbodor:
This simplifies the set up required for the e2e-tests container to open a
window  when docker is running inside VirtualBox on Windows with X11
forwarding.


oehm-smith:
:+1:

PR #501, opened 2017-06-19
Return empty observable instead of emitting null

lbodor:
Returning `empty` has the advantage of not emitting a value when there
isn't one, which is preferable to emitting null. The observable is
terminated normally.


frankfu03:
:+1:

PR #506, opened 2017-06-19
GEOD-514 - Convert to using Page Objects in our e2e tests

oehm-smith:
* SelectSite for this one

### DIFF
--- a/src/e2e/specs/select-site.pageobject.ts
+++ b/src/e2e/specs/select-site.pageobject.ts
@@ -1,0 +1,7 @@
+import { element, by, ElementFinder } from 'protractor';
+
+export class SelectSitePage {
+    readonly url:string = '/';
+    readonly searchBox: ElementFinder = element(by.css('sd-select-site input[name="searchText"]'));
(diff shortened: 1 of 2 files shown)


oehm-smith:
Hold off from reviewing.

src/e2e/specs/select-site.pageobject.ts
```diff
@@ -0,0 +1,9 @@
+import { element, by, ElementHelper, ElementFinder } from 'protractor';
+
+export class SelectSitePage {
+    readonly url:string = '/';
+    readonly searchBox: ElementFinder = element(by.css('sd-select-site input[name="searchText"]'));
```
lbodor: Is `by.binding` an option here? Some guidelines say we should prefer protractor locators. See: https://github.com/CarmenPopoviciu/protractor-styleguide#rule-07-prefer-protractor-locators-when-possible.

oehm-smith: Nope no bindings on this page.

lbodor:
:+1:

PR #507, opened 2017-06-19
Refactor form inputs

frankfu03:


src/client/app/shared/dynamic-form-fields/abstract-gnss-controls.ts
```diff
@@ -1,84 +1,77 @@
...
+    protected isValid(): boolean {
+        return this.formControl && this.formControl.valid;
     }
 
-    protected isInvalid(controlId: string): boolean {
-        let valid: boolean = this.isValid(controlId);
-        return !valid;
+    protected isInvalid(): boolean {
+        return this.formControl && this.formControl.invalid;
     }
```
lbodor: Good. I suspect that in angular !valid does not mean invalid. For example, before a form is validated, it is neither valid nor invalid.

src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
```diff
@@ -24,27 +25,21 @@ export class DatetimeInputComponent extends AbstractGnssControls implements OnIn
...
-    @Input() public label: string = '';
-
     private hours: number = 0;
     private minutes: number = 0;
     private seconds: number = 0;
     private datetimeLast: string = '';
 
     constructor(private elemRef: ElementRef) {
         super();
+        this.required = true;  // default is true for datetime inputs if not specified
```
lbodor: Unnecessary comment, the code speaks for itself.

src/client/app/shared/dynamic-form-fields/number-input.component.html
```diff
@@ -1,18 +1,18 @@
...
          <input (keyup.enter)="ignoreEvent($event);"
              type="number"
              [formControlName]="controlName"
              class="form-control"
              min="{{min}}"
              max="{{max}}"
              step="{{step}}"/>
-         <small>{{getErrorReport(controlName)}}</small>
-
+        <small [hidden]="!formControl.errors" class="alert alert-danger">
```
lbodor: Would it be better to use the `isValid` method defined on the abstract component?

lbodor:
:+1:

src/client/app/shared/dynamic-form-fields/datetime-input.component.ts
frankfu03: Yes, comment line is removed.

src/client/app/shared/dynamic-form-fields/number-input.component.html
frankfu03: In most cases, both are the same Boolean values. If there are error, the form is invalid. But if the form is invalid, there may be no error message. So it may be safe to use "formControl.errors".

lbodor: OK.

PR #509, opened 2017-06-20
GEOD-514 - PageObject update for Select Site w more tests

oehm-smith:
* Created a base class for pages

### DIFF
--- a/src/e2e/specs/base.pageobject.ts
+++ b/src/e2e/specs/base.pageobject.ts
@@ -1,0 +1,35 @@
+import { ElementArrayFinder, ElementFinder } from 'protractor';
+
+export class BasePage {
+
+    /**
+     * Search for the given searchText in any element's text (content) in the array of elements.  An array is
+     * returned from Protractors `element.all(selector)`
+     *
+     * @param elementArray
+     * @param searchText
+     * @return {ElementArrayFinder} of the elements in the array with the searchText
+     */
+    public elementArrayContaining(elementArray: ElementArrayFinder, searchText: string): ElementArrayFinder {
+        return elementArray.filter(function (elem) {
+            return elem.getText().then((text) => {
+                return text === searchText;
+            });
+        });
+    }
+
+    public debug(element: ElementFinder) {
+        element.getInnerHtml().then(
+            (successVal) => {
+                console.log('BasePage.debug success: ', successVal);
+            },
+            (failureVal) => {
+                console.log('BasePage.debug failure: ', failureVal);
+            }
+        );
+    }
+
+    public debugArray(elements: ElementArrayFinder) {
+        elements.each(this.debug);
+    }
+}
--- a/src/e2e/specs/select-site.component.e2e-spec.ts
+++ b/src/e2e/specs/select-site.component.e2e-spec.ts
@@ -8,11 +8,27 @@ describe('SelectSite', () => {
         return await browser.get(selectSitePage.url);
     });
 
-    it('should have an input for siteName/fourCharacterId', () => {
-        expect(selectSitePage.searchBox.isPresent()).toEqual(true);
+    it('entering search text should return list of sites - one site', () => {
+        selectSitePage.searchFor('ade1');
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);
+        expect(selectSitePage.selectSiteListItems.count()).toEqual(1);
+        expect(selectSitePage.selectSiteListItems.first().getText()).toBe('ADE1');
+        selectSitePage.debug(selectSitePage.selectSiteList);
     });
 
-    it('should have a search button', () => {
-        expect(selectSitePage.searchButton.isPresent()).toEqual(true);
+    it('entering search text but do not click on button should return list of sites - one site', () => {
+        selectSitePage.enterSearchText('ade1');
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);
+        expect(selectSitePage.selectSiteListItems.count()).toEqual(1);
+        expect(selectSitePage.selectSiteListItems.first().getText()).toBe('ADE1');
+        selectSitePage.debug(selectSitePage.selectSiteList);
+    });
+
+    it('entering search text should return the specific sites - multiple sites', () => {
+        selectSitePage.searchFor('ade');
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);
+        expect(selectSitePage.selectSiteListItems.count()).toBeGreaterThan(1);
+        expect(selectSitePage.elementArrayContaining(selectSitePage.selectSiteListItems, 'ADE1').count()).toBe(1);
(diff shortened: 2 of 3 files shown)


src/e2e/specs/base.pageobject.ts
```diff
@@ -0,0 +1,36 @@
+import { ElementArrayFinder, ElementFinder } from 'protractor';
+
+export class BasePage {
+
+    /**
+     * Search for the given searchText in any elment's text (content) in the array of elements.  An array is
```
lbodor: spelling: element

src/e2e/specs/base.pageobject.ts
```diff
@@ -0,0 +1,36 @@
...
+    public elementArrayContaining(elementArray: ElementArrayFinder, searchText: string): ElementArrayFinder {
+        return elementArray.filter(function (elem) {
+            return elem.getText().then((text) => {
+                // console.log('array elements text: ', text);
+                return text === searchText;
+            });
+        });
+    }
+
+    public debug(theElement: ElementFinder) {
```
lbodor: Spelling, no articles on identifier names: element

src/e2e/specs/select-site.component.e2e-spec.ts
```diff
@@ -15,4 +15,29 @@ describe('SelectSite', () => {
     it('should have a search button', () => {
         expect(selectSitePage.searchButton.isPresent()).toEqual(true);
     });
```
lbodor: This test is not needed, because the next test will fail if the search buttons is not present.

src/e2e/specs/select-site.component.e2e-spec.ts
```diff
@@ -15,4 +15,29 @@ describe('SelectSite', () => {
...
+        selectSitePage.searchFor('ade1');
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);
+        expect(selectSitePage.selectSiteListItems.count()).toEqual(1);
+        expect(selectSitePage.selectSiteListItems.first().getText()).toBe('ADE1');
+        selectSitePage.debug(selectSitePage.selectSiteList);
+    });
+
+    it('entering search text but dont click on button should return list of sites - one site', () => {
+        selectSitePage.searchFor('ade1');
+        browser.waitForAngular();
```
lbodor: Should `waitForAngular` be done inside `searchFor`?

src/e2e/specs/base.pageobject.ts
```diff
@@ -0,0 +1,36 @@
...
+     * returned from Protractors `element.all(selector)`
+     *
+     * @param elementArray
+     * @param searchText
+     * @return {ElementArrayFinder} of the elements in the array with the searchText
+     */
+    public elementArrayContaining(elementArray: ElementArrayFinder, searchText: string): ElementArrayFinder {
+        return elementArray.filter(function (elem) {
+            return elem.getText().then((text) => {
+                // console.log('array elements text: ', text);
```
lbodor: Remove commented out code.

src/e2e/specs/base.pageobject.ts
```diff
@@ -0,0 +1,36 @@
...
+                console.log('BasePage.debug success: ', successVal);
+            },
+            (failureVal) => {
+                console.log('BasePage.debug failure: ', failureVal);
+            }
+        );
+    }
+
+    public debugArray(theElements: ElementArrayFinder) {
+        theElements.each((theElement: ElementFinder) => this.debug(theElement));
```
lbodor: Consider eta-reduction: `elements.each(this.debug);`

src/e2e/specs/select-site.component.e2e-spec.ts
```diff
@@ -15,4 +15,29 @@ describe('SelectSite', () => {
...
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);
+        expect(selectSitePage.selectSiteListItems.count()).toEqual(1);
+        expect(selectSitePage.selectSiteListItems.first().getText()).toBe('ADE1');
+        selectSitePage.debug(selectSitePage.selectSiteList);
+    });
+
+    it('entering search text should return the specific sites - multiple sites', () => {
+        selectSitePage.searchFor('ade');
+        expect(selectSitePage.selectSiteList.isPresent()).toEqual(true);   //toContain('ADE1');
+        expect(selectSitePage.selectSiteListItems.count()).toBeGreaterThan(1);
```
lbodor: Remove commented code.

src/e2e/specs/select-site.pageobject.ts
```diff
@@ -1,7 +1,20 @@
...
+import { BasePage } from './base.pageobject';
 
-export class SelectSitePage {
+export class SelectSitePage extends BasePage {
     readonly url: string = '/';
     readonly searchBox: ElementFinder = element(by.css('sd-select-site input[name="searchText"]'));
     readonly searchButton: ElementFinder = element(by.css('sd-select-site button'));
+    readonly selectSiteList: ElementFinder = element(by.id('select-site-sites-table'));
+    readonly selectSiteListItems: ElementArrayFinder = element.all(by.css('td[name="siteId"'));
+    readonly siteLog: ElementFinder = element(by.cssContainingText('span', 'Site Information'));
```
lbodor: I don't that this span exists on select site page.

src/e2e/specs/select-site.component.e2e-spec.ts
oehm-smith: This test fails without it so this seems to be ok.

lbodor: That's not what I meant. I didn't suggest that we should remove it.  I think that it may be better to put `waitForAngular` inside `seachFor`, so that clients don't always have to call both. Clients call `searchFor` and if that requires waiting on angular, so be it.

oehm-smith: I've moved it inside `enterSearchText()` where it was needed.  Doesn't need to be after the `click()` since Protractor seems to call this implicitly at the correct time.

lbodor: OK, thanks.

src/e2e/specs/select-site.component.e2e-spec.ts
oehm-smith: We discussed and you suggested that failing on the other tests has the advantage of pointing out which Locator failed.  That is true:

    Failures:
        1) SelectSite entering search text should return list of sites - one site
    Message:                                                                                          
        Failed: No element found using locator: By(css selector, sd-select-sitex button)

The `should have a search button` test is equally helpful due to the test name:

    Failures:                                                                                                                  
        1) SelectSite should have a search button
    Message:
        Expected false to equal true.                                                                                          

However there is no need to have a duplication as you suggest so I'll get rid of these simple tests.  Of course they were handy to write in the beginning.

src/e2e/specs/base.pageobject.ts
oehm-smith: Neat!

lbodor:
Could you squash the commits please?

PR #519, opened 2017-06-21
Replace ResponsiblePartyType class with a simple object

frankfu03:


lbodor:
Typescript 2.4+ adds support for string enums. I suggest that we hold off with replacing the current approach until we upgrade our compiler.

The current approach has the advantage over the proposed of permitting a static guarantee that
`@Input() partyType` in `ResponsiblePartyGroupComponent` can be assigned only one of values specified in `ResponsiblePartyType` (assuming the constructor becomes private) . The proposed string type of `@Input() partyType` would permit any string value as party type.

PR #525, opened 2017-06-26
make end date reactive

heyagosper:


src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -68,6 +68,12 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
         setTimeout(() => {
             if (this.isEditable()) {
                 this.itemGroup.enable();
+                // add a listener for changes to the start date field
+                if (this.itemGroup.controls.startDate) {
+                    this.itemGroup.controls.startDate.valueChanges.subscribe(
+                        updatedStartDate => this.updateEndDateOnPreviousItem(updatedStartDate)
```
oehm-smith: You should be able to write `updatedStartDate => this.updateEndDateOnPreviousItem`

heyagosper: is that because there is only one parameter?

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -291,6 +297,19 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
         }
     }
 
+    /**
+     * Updates the end date on the previous item.
+     */
+    private updateEndDateOnPreviousItem(updatedStartDate: string): void {
+        let previousItem = <FormGroup> this.groupArray.at(this.index+1);
```
oehm-smith: I'd guard this with `if (this.groupArray.length >this.index)`

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -291,6 +297,19 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
...
+        if (previousItem && previousItem.controls.hasOwnProperty('endDate')) {
```
oehm-smith: You should just say '... && previousItems.controls['endDate']`.  What you wrote should work and may take a nanosecond less (as it doesn't need to traverse all the way to Object) but we've used the form I've written everywhere else.

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -291,6 +297,19 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
...
+            let endDateControl = previousItem.controls.endDate;
+            if (endDateControl) {
+                endDateControl.setValue(updatedStartDate);
+                endDateControl.markAsDirty();
```
oehm-smith: I know we may have `markAsDirty()` in other places.  Though is this necessary?  As theoretically dirtiness is also set thru programmatic changes.

heyagosper: doesn't work without this

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
oehm-smith: Not working? I thought that was right.  ...

PR #532, opened 2017-06-26
Stop schlepping change tracking fields

lbodor:
The don't belong on the forms, because they are not user input.


oehm-smith:
Isnt this the same as #533 ?

oehm-smith:
I see stacked.  And I :+1: the 2nd of these so therefore this is :+1: !

PR #534, opened 2017-06-27
Add ng2 validation

frankfu03:


src/client/app/shared/form-input/datetime-input.component.ts
```diff
@@ -66,6 +67,7 @@ export class DatetimeInputComponent extends AbstractInput implements OnInit, DoC
             validators.push(Validators.required);
         }
         validators.push(new DatetimeValidator());
+        validators.push(CustomValidators.maxDate(new Date(MiscUtils.getUTCDateTime())));
```
lbodor: The will reject valid dates. Current time is saved when the component is created and used later when the form is validated, i.e., at time of validation, it is no longer the current time. Implementing this check requires a validator that can itself ask for current time.

lbodor:
These are two pull requests in one. Could we please split this PR, and merge the half that validates URLs.

src/client/app/shared/form-input/abstract-input.component.ts
```diff
@@ -29,6 +29,7 @@ export abstract class AbstractInput implements OnInit {
                     errString += ', ';
                 }
                 let error: any = this.formControl.errors[e];
+                console.log('~~~~~ key='+e+'; error='+error);
```
lbodor: Remove debug logging please.

frankfu03: Yeah, it is done.

PR #572, opened 2017-07-03
Remove item controls class

frankfu03:


src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -312,20 +309,4 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
...
-     */
-    private addFields(itemGroup: FormGroup, formControls: ItemControls): void {
-        if (Object.keys(this.itemGroup.controls).length === 0) {
-            for (let control of formControls.itemControls) {
-                let key: any = Object.keys(control)[0];
-                let value: AbstractControl = control[key];
-                itemGroup.addControl(key, value);
-            }
-        }
-    }
```
lbodor: It's great that this method is gone.

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -118,11 +108,10 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
     abstract getItem(): AbstractViewModel;
 
     /**
-     * Return the controls to become the form.
+     * Return default values and form controls used to define the item form.
      *
-     * @return array of AbstractControl objects
      */
-    abstract getFormControls(): ItemControls;
+    abstract getFormControls(): any;
```
lbodor: Instead of returning `any`, why don't we just return a new `FormGroup`. The implementations of `getFormControl` would call `FormBuilder.group`.

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
```diff
@@ -140,9 +129,17 @@ export abstract class AbstractItemComponent extends AbstractBaseComponent implem
...
-        this.addFields(this.itemGroup, this.getFormControls());
+        this.setupForm();
+    }
+
+    /**
+     * set up the item form, insert it into the form array and populate with initial values if any.
+     *
+     */
+    setupForm() {
+        this.itemGroup = this.formBuilder.group(this.getFormControls());
```
lbodor: If `getFormControls` were to return `FormGroup`, then we could just write
```
this.itemGroup = this.getFormControls();
```
and avoid the `any` return type of `getFormControls`.

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
frankfu03: "getFormControls(): any" has been changed to "getItemForm(): FormGroup" as suggested.

src/client/app/shared/abstract-groups-items/abstract-item.component.ts
frankfu03: Changes have been made as suggested.

lbodor:
:+1:

PR #585, opened 2017-07-10
Add a group html template for all components

frankfu03:


lbodor:
In any inheritance hierarchy, it is required that parent components have no knowledge of child components. `abstract-group.template.html` cannot mention concrete abstract item component implementations like `gnss-receiver-item` and `frequency-standard-item`.

It should be possible to add new concrete group and item components without modifying any code in either the abstract group component or the abstract item component.

I do, however, agree with you that there is too much repetition in the various group item template files. Could you please investigate alternative approaches of removing this boilerplate code? What about using directive `ngTemplateOutlet`?

If you need a template that can contain information about concrete group and item components, then perhaps `site-log.component.html` could be a good option.

frankfu03:
It does not work for adding/removing new items because that it could not pass returnEvents($event) method between item and group components when using ngTemplateOutlet directives.

PR #598, opened 2017-07-21
Add e2e test case on add/save/delete Frequency Standard items

frankfu03:


src/e2e/specs/page-objects/frequency-standard.pageobject.ts
```diff
@@ -0,0 +1,16 @@
+import { by, ElementFinder } from 'protractor';
+import { SiteLogPage } from '../page-objects/site-log.pageobject';
+
+export class FrequencyStandardPage extends SiteLogPage {
```
lbodor: Inheritance is not the correct relationship in this case. Frequency standard group component is not a site log page, instead a site log page has many group components, where one of them is the frequency standard group.

I think that what is missing is an abstract LogItemGroup class. SiteLogPage has many LogItemGroups, LogItemGroup can be FrequencyStandardGroup, GnssReceiverGroup, etc.

src/e2e/specs/page-objects/site-log.pageobject.ts
```diff
@@ -1,15 +1,121 @@
...
+    }
+
     public identifyingElement(): ElementFinder {
         return this.siteInformationHeader;
     }
+
+    public checkInputValue(elemFinder: ElementFinder, elemName: string, expectValue: string) {
+        elemFinder.getAttribute('value').then((value: string) => {
+            console.log('Check if ' + elemName + ' is "' + value + '": ' + (expectValue === value));
+            expect(value).toEqual(expectValue);
```
lbodor: There is a separation of concerns between page objects and tests. Page objects provide access to page elements, tests assert. See: https://martinfowler.com/bliki/PageObject.html and https://www.pluralsight.com/guides/software-engineering-best-practices/getting-started-with-page-object-pattern-for-your-selenium-tests#best-practices. Could we please move all check/expect/assert type of logic into the tests.

src/e2e/specs/utils/test.utils.ts
```diff
@@ -56,4 +58,16 @@ export class TestUtils {
...
         return deferred.promise;
     }
+
+    public static getTimeStamp(): string {
+        return moment().utc().format('YYYY-MM-DD HH:mm:ss');
+    }
+
+    public static toKebabCase(txt: string): string {
+        return lodash.kebabCase(txt);
+    }
```
lbodor: Why wrap lodash? Lodash is so useful that sooner or later most modules end up importing it. It is simpler for clients to call lodash directly. Conventionally, lodash is imported qualified with '_', as in `import * as _ from 'lodash'`.

src/e2e/specs/page-objects/frequency-standard.pageobject.ts
frankfu03: It is done. An LogItemGroup class has been added and  FrequencyStandardGroup class extends from it.

src/e2e/specs/page-objects/site-log.pageobject.ts
frankfu03: The checkInputValue() method has been moved to the  LogItemGroup class.

src/e2e/specs/utils/test.utils.ts
frankfu03: Yeah, it is done.

src/e2e/specs/log-groups/frequency-standard-group.e2e-spec.ts
```diff
@@ -0,0 +1,70 @@
+import { browser } from 'protractor';
+import { SelectSitePage } from '../page-objects/select-site.pageobject';
+import { LoginActions } from '../utils/login.actions';
+import { FrequencyStandardGroup } from '../page-objects/frequency-standard-group.pageobject';
+
+describe('Frequency Standard Group Component', () => {
+
+    let selectSitePage: SelectSitePage = new SelectSitePage();
+    let groupPage: FrequencyStandardGroup = new FrequencyStandardGroup();
```
lbodor: It's not an entire page, just an an item group.

src/e2e/specs/log-groups/frequency-standard-group.e2e-spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        browser.driver.manage().window().maximize();
+        return await browser.get(selectSitePage.url);
+    });
+
+    it('expect should be able to add and save new ' + groupPage.itemName + ' item', () => {
+        loginActions.login('user.a', 'gumby123A');
+        browser.waitForAngular();
+        selectSitePage.openSite(siteId);
+        browser.waitForAngular();
+        expect(groupPage.saveSiteLink.isPresent()).toBe(true);
```
lbodor: A group only represents the part of page that corresponds the an item group component. It does not have a save link. Perhaps `selectSitePage.openSite(siteId)` should return a SiteLogPage object.

```
SiteLogPage siteLogPage = selectSitePage.openSite(siteId);
FrequencyStandardGroup frequencyStandardGroup = siteLogPage.getFrequencyStandardGroup();
...
siteLogPage.save();
```

src/e2e/specs/page-objects/log-item-group.pageobject.ts
```diff
@@ -0,0 +1,114 @@
+import { browser, by, element, ElementFinder, ElementArrayFinder } from 'protractor';
+import * as _ from 'lodash';
+import * as moment from 'moment';
+import { BasePage } from '../page-objects/base.pageobject';
+
+export class LogItemGroup extends BasePage {
```
lbodor: LogItemGroup is not a page, so it shouldn't inherit from BasePage.

src/e2e/specs/page-objects/log-item-group.pageobject.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public getTimeStamp(): string {
+        return moment().utc().format('YYYY-MM-DD HH:mm:ss');
+    }
+
+    public checkInputValue(elemFinder: ElementFinder, elemName: string, expectValue: string) {
+        elemFinder.getAttribute('value').then((value: string) => {
+            console.log('Check if ' + elemName + ' is "' + value + '": ' + (expectValue === value));
+            expect(value).toEqual(expectValue);
```
lbodor: All expect statements should be in test classes not page object classes.

src/e2e/specs/log-groups/frequency-standard-group.e2e-spec.ts
frankfu03: "groupPage" has been changed to "itemGroup" as suggested.

src/e2e/specs/page-objects/log-item-group.pageobject.ts
frankfu03: The checkInputValue() and other methods containing expect() statements have been moved to TestUtils or relevant test classes.

src/e2e/specs/page-objects/log-item-group.pageobject.ts
frankfu03: As suggested, LogItemGroup class does not  inherit from BasePage. Instead, SiteLogPage extends from BasePage.

lbodor:
:+1:

PR #600, opened 2017-07-27
Add e2e test cases for GNSS receiver group

frankfu03:


lbodor:
Performing a login action before every test, as we are currently doing in frequency standard and gnss receiver group tests, will unnecessarily slow down the CI process. Heya worked on setting up a test suite for all tests that require authentication. Does that mechanism work? Perhaps we could speed up CI by adding frequency standard and gnss receiver groups tests to Heya's suite or by similarly re-using a single authenticated browser session to perform multiple tests.

lbodor:
:+1:

frankfu03:
Although each item group has a login statement at the beginning, it will be skipped if it has been already logged in previous tests. The login statement is needed when we only test one or more item groups, not running any other tests from logged-in and logged-out packages.

lbodor:
That's perfect.

PR #616, opened 2017-08-07
E2e test site owner

frankfu03:


src/e2e/specs/page-objects/responsible-party-group.pageobject.ts
```diff
@@ -0,0 +1,69 @@
...
+            city: 'Symonston',
+            administrativeArea: 'ACT',
+            postalCode: '2609',
+            country: 'Australia',
+            email: 'Homer.Simpson@ga.gov.au',
+            primaryPhone: '0262499997',
+            secondaryPhone: '0262499998',
+            fax: '0262499999',
+            url: 'http://www.ga.gov.au'
+        };
```
lbodor: I think that test data belongs in a test, rather than this page object. Could we use the responsible party view model class to represent this test data, rather than `any`?

frankfu03: Yes, the test data should be in test files. Not sure if we should import responsible party view model class from client (outside of e2e folder).

I prefer your proposal "b) externalise the test data into a separate file". We may discuss what kind of file it should be and where to put it.

frankfu03: Test data has been moved to a ts file (view-model.ts), which mocks ResponsiblePartyViewModel, and maybe other view model data if needed.

PR #621, opened 2017-08-09
Add e2e test cases for pressure-sensor group

frankfu03:


src/e2e/specs/log-groups/view-model.ts
```diff
@@ -26,3 +27,16 @@ export const mockResponsibleParty: ResponsiblePartyViewModel = {
     deletedReason: null,
     isDeleted: false
 };
+
+export const mockGnssSensor: any = {
```
lbodor: They are not GNSS sensors, but meteorological sensors.

frankfu03: Yeah, GNSS sensors have been renamed to Meteorological sensors. Thanks.

PR #650, opened 2017-09-08
Fix the scroll-into-view error for Site Owner and Data Source groups

frankfu03:


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -111,6 +111,16 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel>
     abstract getNewItemViewModel(): T;
 
     /**
+     * Returns the element name in kebab case converted from item name.
+     *
+     * Eg., "Site Owner" -> "site-owner"
+     */
+    getElementName(): string {
+        let itemName: string = this.getItemName();
+        return !itemName ? '' : _.kebabCase(itemName);
```
lbodor: How do you feel about `return _.kebabCase(itemName) || ''`, given that `kebabCase` is total/safe?

PR #652, opened 2017-09-08
Remapping accuracy fields for meteorological sensors

frankfu03:


frankfu03:
Ok. No need to make any changes to GeodesyML. 

What I have done is to rename some common fields (accuracy) for meteorological sensors in view model. Each of these  meteorological sensors has a accuracy field, although it has different meaning and unit. Their original names in GeodesyML and label names in web UI remain the same.

frankfu03:
The main purpose of the PR is to simplify the e2e test on meteorological sensors.

PR #653, opened 2017-09-11
Site location required fields

frankfu03:


src/client/app/site-information/site-location.component.ts
```diff
@@ -146,6 +146,23 @@ export class SiteLocationComponent extends AbstractBaseComponent implements OnIn
     }
 
     /**
+     * Returns true if not all fields are either empty or valid
+     */
+    public isCartesianPositionRequired(): boolean {
+        let x: number = this.cartesianPositionForm.get('x').value;
+        let y: number = this.cartesianPositionForm.get('y').value;
+        let z: number = this.cartesianPositionForm.get('z').value;
+        return (!!x || x === 0) || (!!y || y === 0) || (!!z || z === 0);
```
lbodor: Consider,
```
return x != null || y != null || z != null;
```
Because of type coercion, sometimes `==` and `!=` are better than their type-safe variants. `undefined == null` is `true`, but `undefined === null` is false.

src/client/app/site-information/site-location.component.ts
```diff
@@ -214,7 +231,7 @@ export class SiteLocationComponent extends AbstractBaseComponent implements OnIn
     private handleLocationPositionGroupChange(groupItems: any, positionGroup: FormGroup, permanentValidators: ValidatorFn[]) {
         let someFieldHasValue: boolean = false;
         Object.keys(groupItems).forEach((key) => {
-            if (groupItems[key]) {
+            if (groupItems[key] || groupItems[key] === 0) {
                 someFieldHasValue = true;
             }
```
lbodor: There is duplication of logic here and in methods `isCartesianPositionRequired` and `isCartesianPositionRequired`. I'm not sure what the best way to resolve it. Perhaps the generic approach in `handleLocationPositionGroupChange` can be factored out and made available to the template.

src/client/app/site-information/site-location.component.ts
frankfu03: Yeah, it works. Better solution. Thanks.

src/client/app/site-information/site-location.component.ts
frankfu03: As suggested, I refactored handleLocationPositionGroupChange () method to return a boolean flag indicating whether any fields have valid values, and removed the two methods isCartesianPositionRequired() and isGeodeticPositionRequired() as they contain duplicate logic.

lbodor: That's great.

PR #655, opened 2017-09-12
Responsible party required fields

frankfu03:


frankfu03:
I have moved flags of isDataType and isMetadataCustodian to getItemForm() method so that they can be initialised early.

PR #667, opened 2018-01-30
Add validation checking if endDate is before startDate

frankfu03:
I think we should rebase this branch onto `master`, and deploy it as a enddate-validation-hotfix-0.1.1. There many new changes on `next` that haven't yet been deployed to production.


src/client/app/shared/form-input/datetime-input.component.ts
```diff
@@ -64,7 +65,17 @@ export class DatetimeInputComponent extends AbstractInput implements OnInit, DoC
...
+        if (this.controlName === 'endDate') {
+            let startDateControl: FormControl = <FormControl>this.form.controls.startDate;
+            validators.push(new DatetimeRangeValidator(startDateControl, true));
+        } else if (this.controlName === 'startDate') {
+            let endDateControl: FormControl = <FormControl>this.form.controls.endDate;
+            validators.push(new DatetimeRangeValidator(endDateControl, false));
+        } else {
+            validators.push(new DatetimeValidator());
+        }
+
```
lbodor: Is it possible to push multiple validators into `validators`? If yes, would that simplify the implementation of `DatetimeRangeValidator`, so that it doesn't have to also do the job of `DatetimeValidator`? Would the first validation error cause other validators in `validators` to be skipped?

src/client/app/shared/form-input-validators/datetime-range-validator.ts
```diff
@@ -0,0 +1,55 @@
...
+                }
+                return {invalid_datetime_format: 'Invalid date (valid format: ' + datetimeFormat + ')'};
+            } else if (!otherDate.isValid()) {
+                return null;
+            }
+
+            this.otherDateControl.setErrors(null);
+            if (this.isValidatingEndDate) {
+                if (thisDate.isSameOrBefore(otherDate)) {
+                    return {invalid_datetime_format: 'Date Removed/End must be after Date Installed/Start'};
```
lbodor: Is there something on `thisDateControl` that could let us choose between displaying "Date Removed" and "End Date"? Perhaps it can be done via the parent of `thisDateControl`, which should have a map of names of its children controls (include the name of `thisDateControl`).

PR #671, opened 2019-02-11
Radio button group

frankfu03:


src/client/app/shared/form-input/radiobuttons-input.component.ts
```diff
@@ -0,0 +1,63 @@
...
+            this.formControl.setValidators(validators);
+        });
+    }
+
+    protected updateChanges() {
+        this.formControl.markAsDirty();
+    }
+
+    protected parseInputOptions(): void {
+        if (!this.formControl.value) {
```
lbodor: Can this happen? Is there a default value?

frankfu03: This is to check if one of the radio buttons is toggled on (no one will be toggled on when the model value is null), also to ensure the model value is one of the option's values, for example, the model value for siteStatus radio buttons must be either "PUBLIC" or "PRIVATE". 

Yes, we can set "PUBLIC" as the default value for siteStatus radio buttons, but this may hide some potential errors in CORS site.

lbodor: I don't think that we should show a screen with neither "open" nor "closed" selected, even if the model says it's null. The default is "open", even if the database says it's null. We can check tomorrow in the stand-up.

frankfu03: OK, I'll set "PUBLIC" as the default value for siteStatus radio buttons.

PR #675, opened 2019-02-17
Site administration tab

frankfu03:
Looks good. Could you please try to get the tests to pass and merge.

### DIFF
--- a/src/client/app/shared/cors-site/cors-site.service.ts
+++ b/src/client/app/shared/cors-site/cors-site.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, OnDestroy } from '@angular/core';
-import { Http, Response } from '@angular/http';
+import { Headers, Http, Response } from '@angular/http';
 import { Observable } from 'rxjs/Rx';
 import { Subject } from 'rxjs/Subject';
 import 'rxjs/add/operator/map';
@@ -88,6 +88,23 @@ export class CorsSiteService implements OnDestroy {
       .catch(HttpUtilsService.handleError);
   }
 
+  saveCorsSite(siteAdminModel: SiteAdministrationModel, id_token: string) : Observable<Response> {
+    console.log('Save existing CorsSite - CorsSiteViewModel: ', siteAdminModel);
+
+    //const user: User = this.authService.user.value.id_token;
+
+    const headers = new Headers();
+    if (id_token) {
+      headers.append('Authorization', 'Bearer ' + id_token);
+    }
+
+    return this.http.post(this.constantsService.getWebServiceURL() + '/corsSites/upload',
+                          this.getGeodesyMlFromViewModel(siteAdminModel),
+                          { headers: headers })
+        .map(HttpUtilsService.handleJsonData)
+        .catch(HttpUtilsService.handleError);
+  }
+
   private fixWFSeData(wfsData: any): any {
     // TODO - make this data an Interface
     console.debug('cors-site service - from wfsService - fixWFSeData - data: ', wfsData);
@@ -98,6 +115,10 @@ export class CorsSiteService implements OnDestroy {
     });
     console.debug('cors-site service - from wfsService - fixWFSeData - return: ', fieldsDefined);
     return fieldsDefined; //data;
+  }
+
+  private getGeodesyMlFromViewModel(siteAdminModel: SiteAdministrationModel): string {
+    return '';
   }
 }
 
--- a/src/client/app/shared/cors-site/cors-site.service.ts
+++ b/src/client/app/shared/cors-site/cors-site.service.ts
@@ -18,7 +18,7 @@ import { SiteAdministrationModel } from '../../site-administration/site-administ
 @Injectable()
 export class CorsSiteService implements OnDestroy {
 
-  public isSuperUser: BehaviorSubject<boolean> = new BehaviorSubject(false);
+  public isSuperuser: BehaviorSubject<boolean> = new BehaviorSubject(false);
   private unsubscribe: Subject<void> = new Subject<void>();
 
   /**
--- a/src/client/app/shared/cors-site/cors-site.service.ts
+++ b/src/client/app/shared/cors-site/cors-site.service.ts
@@ -2,6 +2,9 @@ import { Injectable, OnDestroy } from '@angular/core';
 import { Headers, Http, Response } from '@angular/http';
 import { Observable } from 'rxjs/Rx';
 import { Subject } from 'rxjs/Subject';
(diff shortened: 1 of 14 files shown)


src/client/app/shared/cors-site/cors-site.service.ts
```diff
@@ -43,6 +60,26 @@ export class CorsSiteService {
         });
   }
 
+  /**
+   * Returns an Observable for the HTTP GET request for the REST Web Service resource.
+   * @param {string} siteId - The Four Character Id of the site.
+   * @return {SiteAdministrationModel} The Observable for the HTTP request.
+   */
+  getSiteAdministration(siteId: string): Observable<SiteAdministrationModel> {
+    let url = this.constantsService.getWebServiceURL() + '/corsSites/search/findByFourCharacterId?id=' + siteId;
```
lbodor: Could we instead use `/corsSites?fourCharacterId=`, which we get for free from spring data rest, as opposed to search/findByFourCharacterId, which we have implemented ourselves, and could consider removing one day.

src/client/app/shared/global/user-auth.service.ts
```diff
@@ -86,6 +86,22 @@ export class UserAuthService {
         }
     }
 
+    public isSuperUser(): Observable<boolean> {
```
lbodor: isSuperuser, lower case s, like below

src/client/app/shared/global/user-auth.service.ts
```diff
@@ -86,6 +86,22 @@ export class UserAuthService {
...
+        }
+
+        let isSuperuser = false;
+        for (let auth of this.user.value.profile.authorities) {
+            if (auth.toLowerCase() === 'superuser') {
+                isSuperuser = true;
+            }
+        }
+
+        console.log(this.user.value.profile.sub +' is ' + (isSuperuser ? '' : 'not ') + 'a super user');
```
lbodor: I think we should remove this log statement, because it's a somewhat ad-hoc. We could instead, in this PR or later, log the entire user profile to the console at the time of login.

src/client/app/shared/global/user-auth.service.ts
lbodor: What's the reason for this method returning an observable? Could it just return a boolean instead?

src/client/app/site-administration/site-administration-model.ts
```diff
@@ -0,0 +1,13 @@
+/**
+* This is the Data & View Models for Site Administration.
+*/
+
+export class SiteAdministrationModel {
+    public id: number;
+    public siteStatus: string;
+
+    constructor(id: number = null, status: string = 'PUBLIC') {
```
lbodor: Is the null default value for the id field ever used?

src/client/app/site-administration/site-administration.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    public sitePermissionList: RadioButtonOption[] = [
+        {
+            label: 'Open Data',
+            value: 'PUBLIC',
+            tips: 'Access to data from these stations is open to anyone for access, use or sharing.'
+        },
+        {
+            label: 'Closed Data',
+            value: 'PRIVATE',
+            tips: 'Access to data from these stations is restricted to APREF analysis centres'
```
lbodor: I think public and open data should be enumerations (value, name, description), so that we don't have to repeat strings 'PUBLIC' and 'PRIVATE', but can instead refer to enum values.

src/client/app/site-log/site-log.component.ts
```diff
@@ -237,6 +250,44 @@ export class SiteLogComponent implements OnInit, OnDestroy {
...
+    public saveExistingCorsSite() {
+        if (!this.siteLogService.isUserAuthorisedToEditSite.value || !this.corsSiteForm.dirty) {
+            this.isLoading = false;
+            return;
+        }
+
+        let newSiteStatus = this.corsSiteForm.getRawValue().siteAdministration.siteStatus;
+        if (this.siteAdminModel.siteStatus === newSiteStatus) {
+            this.isLoading = false;
+            this.dialogService.showLogMessage('Skip updating siteStatus in CORS Site as no changes have been made for '
```
lbodor: This seems like a debugging message that the user probably shouldn't see.

src/client/app/site-log/site-log.component.ts
```diff
@@ -347,6 +419,16 @@ export class SiteLogComponent implements OnInit, OnDestroy {
                         this.siteLogForm.disable();
                     }
                 });
+
+                this.userAuthService.isSuperUser().subscribe(superuser => {
```
lbodor: Does isSuperuser need to be an observable? hasAuthorityToEditSite does because it performs http to get the answer, but isSuperuser does not.

src/client/app/shared/cors-site/cors-site.service.ts
frankfu03: OK, it is done.

src/client/app/shared/global/user-auth.service.ts
frankfu03: Yeah, changed from isSuperUser to isSuperuser.

src/client/app/shared/global/user-auth.service.ts
frankfu03: OK, removed.

src/client/app/shared/global/user-auth.service.ts
frankfu03: Yes, I changed to return a boolean and it works fine. I may do the same change to public hasAuthorityToEditSite(siteId: string): Observable<boolean>{} in a separate PR later.

src/client/app/site-administration/site-administration-model.ts
frankfu03: Yes, it was used in PrefetchCorsSiteResolver class when creating a newSite.

        _if (fourCharacterId === 'newSite') {
            return Observable.of(new SiteAdministrationModel());
        }_

src/client/app/site-administration/site-administration.component.ts
frankfu03: Good suggestion. I will do it in separate PR later.

src/client/app/site-log/site-log.component.ts
frankfu03: Ok, changed to console.debug().

src/client/app/site-log/site-log.component.ts
frankfu03: Changed to return a boolean as did above.

PR #680, opened 2019-02-27
Network http service

frankfu03:


src/client/app/shared/cors-network/cors-network.service.ts
```diff
@@ -0,0 +1,107 @@
...
+            .catch(HttpUtilsService.handleError);
+    }
+
+    /**
+     * Returns an Observable for the HTTP GET request for the REST Web Service resource.
+     * @param {string} networkName - The name of the network.
+     * @return {CorsNetworkModel} The Observable of a CORS network object for the HTTP request.
+     */
+    getCorsNetworksByName(networkName: string): Observable<CorsNetworkModel> {
+        return this.http.get(this.webServiceUrl + '/search/findByName?name=' + networkName)
```
lbodor: Prefer `/corsNetworks?name=` to `/corsNetwork/search/findByName?name=`

src/client/app/shared/cors-network/cors-network.service.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Recursively retrieve all CORS networks from each page of the HTTP request
+     *
+     * @param {CorsNetworkModel[]} corsNetworks - An array storing all CORS networks retrieved
+     * @param {string} url - The URL of the HTTP request with page number parameter
+     * @return {CorsNetworkModel[]} The Observable of a list of CORS networks for the HTTP request.
+     */
+    private getCorsNetworksFromPage(corsNetworks: CorsNetworkModel[], url: string): Observable<CorsNetworkModel[]> {
```
lbodor: You should be able to write a type-generic function here: `private getResourcesFromPage<T>(resources: T[], url: string, resourceParser: (item: any) => T): Observable<T[]>`.

You would then be able to call this function to fetch all CorsNetworks as well as all CorsSites (as in your previous PR).

src/client/app/shared/cors-network/cors-network.service.ts
frankfu03: Ok, changed to `/corsNetwork?name=`.

src/client/app/shared/cors-network/cors-network.service.ts
```diff
@@ -0,0 +1,107 @@
...
+import { ConstantsService } from '../global/constants.service';
+import { CorsNetworkModel } from './cors-network-model';
+
+/**
+ * This class provides the service with methods to retrieve CORS networks from DB.
+ */
+@Injectable()
+export class CorsNetworkService {
+
+    private webServiceUrl: string;
```
lbodor: I think that resourceUrl would be a better name. https://gws.geodesy.ga.gov.au is the API/web service URL, but since corsNetworks are resources, https://gws.geodesy.ga.gov.au/corsNetworks is a resourceUrl.

src/client/app/shared/cors-network/cors-network-model.ts
```diff
@@ -0,0 +1,14 @@
+/**
+* This is the Data & View Models for CORS Network.
+*/
+export class CorsNetworkModel {
+    public id: number;
+    public name: string;
+    public description: string;
+
+    constructor(id: number = null, name: string = null, description: string = null) {
```
lbodor: Would it be possible to disallow these nulls?

src/client/app/shared/cors-network/cors-network.service.ts
frankfu03: Good. I've added a type-generic function to HttpUtilsService class so that it can be used by both CorsNetworks and CorsSites classes.

src/client/app/shared/global/http-utils.service.ts
```diff
@@ -50,4 +50,30 @@ export class HttpUtilsService {
...
+     *
+     * @param {T[]} resources - An array of all resources retrieved
+     * @param {string} url - The URL of the HTTP request with page number parameter
+     * @return {T[]} The Observable of a list of resources for the HTTP request.
+     */
+    public getResourcesFromPage<T>(resources: T[], url: string, resourceParser: (item: any) => T): Observable<T[]> {
+        return this.http.get(url).pipe(
+            mergeMap((response: Response) => {
+                let data = response.json();
+                let items: any[] = data ? data['_embedded']['corsNetworks'] : [];
```
lbodor: oops, can't have corsNetworks here

lbodor: Perhaps you could grab the only child of _embedded, there seems to be only one.

src/client/app/shared/cors-network/cors-network.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Returns an Observable for the HTTP GET request for all records available from the CORS Network DB.
+     *
+     * @return {CorsNetworkModel[]} The Observable of a list of CORS networks for the HTTP request.
+     */
+    getAllCorsNetworks(): Observable<CorsNetworkModel[]> {
+        let url = this.webServiceUrl + '?size=1000';
+        return this.httpUtils.getResourcesFromPage([], url, (item: any): CorsNetworkModel => {
+            return new CorsNetworkModel(item.id, item.name, item.description);
```
lbodor: This looks good now, it's so much shorter.

src/client/app/shared/cors-network/cors-network-model.ts
frankfu03: Yeah, removed default values null from CorsNetworkModel constructor.

src/client/app/shared/cors-network/cors-network.service.ts
frankfu03: Yeah, changed to resourceUrl.

src/client/app/shared/global/http-utils.service.ts
frankfu03: How about add another parameter to the function: 
`getResourcesFromPage<T>(resourceName: string, resources: T[], url: string, resourceParser: (item: any) => T): Observable<T[]>` ?

When calling it in CorsNetworkService: 
` let resourceName = 'corsNetworks';`

lbodor: I think we should derive it if we can. Would this work:

embeded = data['embedded']
items = embedded[Object.keys(embedded)[0]]

lbodor: or just
items = Object.values(data['embedded'])[0]

frankfu03: Thanks. I used your first method:

                   ```
 let firstKey = Object.keys(data['_embedded'])[0];
  items = data['_embedded'][firstKey];
```

`Object.values(data['embedded'])` does not work: _Property 'values' does not exist on type 'ObjectConstructor'_.

lbodor: ok

PR #681, opened 2019-02-27
Hotfix prefetch cors network resolver

frankfu03:
Importing files from index.ts caused a error when loading webpage:

`Error: Can't resolve all parameters for PrefetchCorsNetworkResolver: ([object Object], ?, ?)`

This is likely a circular dependency issue resulting from importing files from index.ts. It is a good practice to import each file separately. Angular CLI wouldn't have this problem.


frankfu03:
Should merge into master, not next.

PR #683, opened 2019-03-20
Cors networks

frankfu03:


src/client/app/shared/cors-site/cors-site.service.ts
```diff
@@ -72,10 +72,16 @@ export class CorsSiteService implements OnDestroy {
       .map((response: Response) => {
         if (response.status === 200) {
           let data: any = response.json();
-          return new SiteAdministrationModel(data.id, data.siteStatus);
+          let corsSite = data && data['_embedded'] ? data['_embedded']['corsSites'][0] : null;
+          if (corsSite) {
+              let siteAdmin = new SiteAdministrationModel(corsSite.id, corsSite.siteStatus);
+              siteAdmin.parseNetworkTenancies(corsSite.networkTenancies);
```
lbodor: Could the SiteAdministrationModel constructor parse network tenancies, so that this extra step is not needed?

lbodor: new SiteAdministrationModel(corsSite)

src/client/app/shared/cors-site/cors-site.service.ts
```diff
@@ -72,10 +72,16 @@ export class CorsSiteService implements OnDestroy {
       .map((response: Response) => {
         if (response.status === 200) {
           let data: any = response.json();
-          return new SiteAdministrationModel(data.id, data.siteStatus);
+          let corsSite = data && data['_embedded'] ? data['_embedded']['corsSites'][0] : null;
```
lbodor: I think we should be more defensive here and also issue an error if there are more than one corsSites in the result set, so that we don't potentially one day modify the wrong one.

src/client/app/shared/cors-site/cors-site.service.ts
```diff
@@ -96,17 +102,29 @@ export class CorsSiteService implements OnDestroy {
...
 
-  saveCorsSite(siteAdminModel: SiteAdministrationModel, id_token: string) : Observable<any> {
-    console.log('Save existing CorsSite - SiteAdminModel: ', siteAdminModel);
+  updateCorsSite(siteAdminModel: SiteAdministrationModel): Observable<any> {
+    let url = this.constantsService.getWebServiceURL() + '/corsSites/' + siteAdminModel.id;
+    return this.http.patch(url, siteAdminModel, {headers: this.getHttpHeaders()});
+  }
+
+  addToNetwork(siteId: number, networkId: number): Observable<any> {
+    let url = this.constantsService.getWebServiceURL() + '/corsSites/' + siteId + '/addToNetwork?networkId=' + networkId;
```
lbodor: A corsSite json response includes links to addToNetwork and removeFromNetwork, could we use those values instead of reconstructing the URLs here? The server decides.

src/client/app/site-administration/site-administration-model.ts
```diff
@@ -1,13 +1,47 @@
...
 
     constructor(id: number = null, status: string = 'PUBLIC') {
         this.id = id;
         this.siteStatus = status;
+        this.corsNetworks = [];
+    }
+
+    public parseNetworkTenancies(networkTenancies: any[]) {
+        if (networkTenancies) {
+            networkTenancies.forEach((network: any) => {
```
lbodor: Prefer higher-order functions to for-loops. In this case you could use `map`.

src/client/app/site-log/site-log.component.ts
```diff
@@ -487,6 +459,91 @@ export class SiteLogComponent implements OnInit, OnDestroy {
...
+            siteAdminUpdateList.push( {action: 'removing', networkId: network.id} );
+        });
+        return siteAdminUpdateList;
+    }
+
+    /*
+     * Update CORS site properties by recursively/sequentially invoking HTTP client service for each property changed
+     *
+     */
+    private updateCorsSiteProperties(siteAdminUpdateList: any[], index: number) {
```
lbodor: Naming, generally you should avoid suffixing variable names with `List`, instead prefer the shorter siteAdminUpdates.

src/client/app/shared/cors-site/cors-site.service.ts
frankfu03: Yeah, done as suggested.

src/client/app/shared/cors-site/cors-site.service.ts
frankfu03: Good. I modified the code so that it throws an error if more than 1 corsSite found.

src/client/app/shared/cors-site/cors-site.service.ts
frankfu03: It is done: links to _addToNetwork_ and _removeFromNetwork_ are retrieved from corsSite json response, and the two functions _addToNetwork()_ and _removeFromNetwork()_ are megered into one: _updateNetwork(href: string, networkId: number)._

src/client/app/site-log/site-log.component.ts
frankfu03: Yeah, changed to plural variables.

src/client/app/site-administration/site-administration-model.ts
frankfu03: Done.

frankfu03:
Can we switch off e2e tests for now. We can switch it on once we fix it.

lbodor:
I think we should leave them on and let them fail. It could happen that they start failing for another reason, which would be a sign of a new bug.

frankfu03:
OK.

src/client/app/shared/form-input/multi-select-boxes.component.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    public selects: CorsNetworkModel[];
+
+    propagateChange: Function = (_: any) => { };
+    propagateTouch: Function = () => { };
+
+    constructor() {
+        super();
+        this.selects = [];
+    }
```
lbodor: Would this constructor be needed if we were to change the assignment to an initialiser?

```
public selects: corsNetworkModel[] = [];
```

src/client/app/shared/form-input/multi-select-boxes.component.ts
```diff
@@ -0,0 +1,123 @@
...
+     * Initialise all data on loading the site-log page
+     */
+    public ngOnInit() {
+        super.ngOnInit();
+        this.selects = this.formControl ? this.formControl.value : [];
+    }
+
+    canAddItems(): boolean {
+        if (this.form.disabled)
+            return false;
```
lbodor: Curly braces:
```
if (this.form.disabled) {
  return false;
}
```

src/client/app/shared/form-input/multi-select-boxes.component.ts
```diff
@@ -0,0 +1,123 @@
...
+        return this.hasItemsSelected(this.options);
+    }
+
+    canRemoveItems(): boolean {
+        if (this.form.disabled)
+            return false;
```
lbodor: braces

src/client/app/shared/form-input/multi-select-boxes.component.ts
```diff
@@ -0,0 +1,123 @@
...
+        return this.hasItemsSelected(this.selects);
+    }
+
+    addSelectedItems() {
+        if (this.form.disabled)
+            return;
```
lbodor: braces, If you would like to omit braces you could put everything on one line:
```
if (this.form.disabled) return;
```

src/client/app/site-administration/site-administration-model.ts
```diff
@@ -1,13 +1,52 @@
...
+ * This is the Data & View Models for Site Administration.
+ */
 export class SiteAdministrationModel {
     public id: number;
     public siteStatus: string;
+    public addToNetworkHref: string;
+    public removeFromNetworkHref: string;
+    public corsNetworks: CorsNetworkModel[] = [];
+
+    constructor(corsSite: any = null) {
```
lbodor: It would be great to read up a little on some more advanced type techniques, so that we can give a type to `corsSite` other than `any`. The reasoning is that we know that this method will fail if it is invoked with a string or a number, and the types should say that. I know that the parameter value comes from a back-end service, but what options do we have?

src/client/app/site-administration/site-administration-model.ts
```diff
@@ -1,13 +1,52 @@
...
+            }
+        });
+    }
+
+    private findCorsNetworkModelById(networkId: number, corsNetworkList: CorsNetworkModel[]): CorsNetworkModel {
+        for (let network of corsNetworkList) {
+            if (network.id === networkId) {
+                return network;
+            }
+        }
```
lbodor: Prefer high-order combinators to loops, consider _find from lodash.
```
_find(corsNetworks, network => network.id === networkId)
```
or
```
_find(corsNetworks, ['id', networkId])
```
This is now so short that you don't need `findCorsNetworkModelById`

src/client/app/site-log/site-log.component.ts
```diff
@@ -487,6 +459,72 @@ export class SiteLogComponent implements OnInit, OnDestroy {
...
+        });
+
+        return observables;
+    }
+
+    /*
+     * Update CORS site properties by recursively/sequentially invoking HTTP client service for each property to be changed
+     *
+     */
+    private updateCorsSiteProperties(observables: Observable<Response>[], index: number) {
```
lbodor: It is odd to see recursion and indexing used together like this, the index is unnecessary. Consider using structural recursion:
```updateCorsSiteProperties(observables: Observable<Response>[] {
    if (observables) { // base case
       ...
    } else { // recursive case
       ...
       // use head: _head(observables)
       // and recurse on tail: updateCorsSiteProperties(_tail(observables))
}
```
There could also be a solution using `_reduce` from lodash.

src/client/app/site-administration/site-administration-model.ts
lbodor: For example, see https://stackoverflow.com/questions/51990839/how-do-i-map-http-request-response-to-my-defined-object-in-typescript.

Is our version of http.get paremetric?

src/client/app/shared/form-input/multi-select-boxes.component.ts
frankfu03: Ok, the line "this.selects = []" has been removed.

src/client/app/shared/form-input/multi-select-boxes.component.ts
frankfu03: Done.

src/client/app/shared/form-input/multi-select-boxes.component.ts
lbodor: Can the constructor be removed?

src/client/app/shared/form-input/multi-select-boxes.component.ts
frankfu03: Yeah, done.

src/client/app/shared/form-input/multi-select-boxes.component.ts
frankfu03: Done.

src/client/app/site-administration/site-administration-model.ts
frankfu03: Used lodash _.find(corsNetworks, ['id', networkId]) as suggested. Thanks.

src/client/app/site-log/site-log.component.ts
frankfu03: I used _.head() and _.tail() as suggested. It is better.

src/client/app/shared/form-input/multi-select-boxes.component.ts
frankfu03: Yes, it can be removed as its parent class AbstractInput does not have a constructor.

src/client/app/site-administration/site-administration-model.ts
frankfu03: I am not sure if  our version of http.get(0 is paremetric.

I have added 5 interfaces to src/client/app/shared/cors-site/cors-site-model.ts to replace the "any" types.

PR #692, opened 2019-04-17
Misc ui fixes

frankfu03:


tools/env/dev.ts
```diff
@@ -8,7 +8,7 @@ const DevConfig: EnvConfig = {
   CLIENT_URL: 'https://dev.gnss-site-manager.geodesy.ga.gov.au',
   WEB_URL_LINKS: [
     {name: 'GNSS Data Repository', url: 'https://dev-data.gnss.ga.gov.au'},
-    {name: 'GA Regional GNSS Datacentre', url: 'https://www.auscors.ga.gov.au'},
+    {name: 'AUSCORS', url: 'https://www.auscors.ga.gov.au'},
```
lbodor: How should be order these items? Should AUSCORS go on top to keep GNSS Data Repo and GNSS Portal together?

frankfu03: Yeah, reordered as suggested.

PR #693, opened 2019-04-23
Reformat header & footer

frankfu03:


lbodor:
Is this a new kind of test failure? There is no usual mention of an unexpected alert popup.

frankfu03:
Yes, it was introduced by the rename of the "NEW" buttons in previous merge. 

For example, the error message is: 
       Failed: No element found using locator: by.buttonText("New Collocation Information")

Since the "New Collocation Information" button was renamed to "New", so the e2e test couldn't find the button element. I'll fix them in a new PR.

lbodor:
OK, great. I think we should merge that new PR first and rebase to potentially uncover any new problems with the 2 current PRs.

frankfu03:
The PR for fixing e2e "New" button name issue has been merged.

PR #694, opened 2019-04-26
Add new GNSS receiver item by updating the current one

frankfu03:


src/client/app/gnss-receiver/gnss-receivers-group.component.ts
```diff
@@ -31,4 +34,44 @@ export class GnssReceiversGroupComponent extends AbstractGroupComponent<GnssRece
     getNewItemViewModel(): GnssReceiverViewModel {
         return new GnssReceiverViewModel();
     }
+
+    /**
+     * Add a new GNSS Receiver by copying and updating values from the current one
+     */
+    updateCurrentReceiver(event: UIEvent) {
```
lbodor: I wonder if we can avoid some of the duplication introduced here, like date and event management, if we were to instead first call addNew and then copy all field values from the previous log entry.

frankfu03: Yes, there are some duplication in code. I tried calling addNew() first and then invoking patchValue() by copyinh values from previous item, but failed in Satellite Systems which are a group of checkboxes.

frankfu03: Ok, it is done. Added a listener to checkboxes-input component which will update checkbox fields upon changing value (setValue) for Satellite Systems input.

src/client/app/gnss-receiver/gnss-receivers-group.component.ts
```diff
@@ -31,4 +33,31 @@ export class GnssReceiversGroupComponent extends AbstractGroupComponent<GnssRece
...
+            attrNames.map((attrName: string) => {
+                let value = oldFormGroup.get(attrName).value;
+                if (value) {
+                    newFormGroup.get(attrName).setValue(value);
+                    newFormGroup.get(attrName).markAsDirty();
+                }
+            });
+            newFormGroup.controls.receiverType.disable();
+            newFormGroup.controls.manufacturerSerialNumber.disable();
+        });
```
lbodor: Would it be possible to use patchValue instead?

frankfu03: Yes, we can use patchValue(), but have more work to do. The addNew() function will create a new Receiver view model with most fields empty, but startDate and dateInserted are set with current date/time. To using pachValue(), we have to make a copy of current item values, set null to id and endDate, and make sure both startDate and dateInserted in new item won't be override, etc.

Copying the required field values maybe is a reliable and safe way. It can also mark dirty for fields whose values are not empty and copied.

lbodor: I thought patchValue() was made to update only a subset of values, so perhaps there wouldn't be a problem with preserving the values provided by addNew(). I would just go towards eliminating the loop. You would take a subset of the previous record and use that to patch the same subset of the new record. I do agree with you about markAsDirty(), which would need to be set manually anyway.

frankfu03: OK, it is done. Used patchValue() to update a subset of values picked from previous receiver.

lbodor: :+1:

PR #695, opened 2019-04-30
Fix button name errors in e2e test

frankfu03:


lbodor:
What other errors does the second commit fix? The old-standing "alert" error?

frankfu03:
Yes, we still have "alert" errors plus timeout errors. I have dropped the 2nd commit and put it to a separate PR.

src/client/app/site-information/site-location.component.html
```diff
@@ -18,6 +18,7 @@
         </span>
         <span [hidden]="siteLocation != null" class="pull-right">
             <button class="btn btn-similar btn-xs"
+                    id="new-{{getElementName()}}"
                     title="Add a new {{getItemName()}}"
```
lbodor: Do you think that we could at some stage remove new and delete buttons from site location?

frankfu03: Yes.

PR #697, opened 2019-05-03
Add "Edit" buttons to enable edit site log items

frankfu03:


src/client/app/shared/abstract-groups-items/abstract-base.component.ts
```diff
@@ -5,9 +5,12 @@ import { SiteLogService } from '../site-log/site-log.service';
 export abstract class AbstractBaseComponent implements OnDestroy {
 
     public isEditable: boolean;
+    public isItemEditable: boolean;
```
lbodor: Attributes isEditable and isItemEditable are too similar.

src/e2e/specs/page-objects/log-item-group.pageobject.ts
```diff
@@ -56,6 +56,11 @@ export class LogItemGroup {
         return itemContainer.element(by.buttonText('Delete'));;
     }
 
+    public getEditButton(): ElementFinder {
+        let itemContainer: ElementFinder = this.getItemContainer(this.newItemIndex);
+        return itemContainer.element(by.buttonText('Edit'));;
```
lbodor: too many semicolons

src/client/app/shared/abstract-groups-items/abstract-base.component.ts
frankfu03: Yes, I do it on purpose as isEditable should be isAuthorised in fact. I may do it in a separate PR.

src/e2e/specs/page-objects/log-item-group.pageobject.ts
frankfu03: yeah, extra semicolon is removed.

frankfu03:
If "revert" is to discard the changes and changes the fields back to read-only, wouldn't "cancel" be more appropriate?

lbodor:
I agree, cancel would be better.

frankfu03:
This PR has too much changes and rollbacks. I would like to close it and open another clean PR.

PR #699, opened 2019-05-21
Refactor compare function

frankfu03:


lbodor:
Could you please split into two PRs? One PR per logical change. A refactoring, by definition, does not change functionality. Perhaps one PR to refactor/simplify compare functions and one to change the sorting of responsible parties.

Also, our use of MiscUtils is getting out of control. What are the criteria of putting something into MiscUtils? I think we should be removing code from MiscUtils, so that we can delete it altogether one day.

frankfu03:
As suggested, I have changed this PR to refactor/simplify compare functions. Another new PR for the sorting of responsible parties is done but will be on hold due to primary/secondary contacts issue.

frankfu03:
This PR is ready for further review.

frankfu03:
Thanks. I may do these changes later.

PR #703, opened 2019-06-03
Convert antenna and receiver types to upper case

lbodor:


lbodor:
This is a quick fix to guard against people entering antenna and receiver types in lowercase. It should be superseded by our coming work on antenna and receiver type drop-downs.

PR #706, opened 2019-06-05
Refactor addNewByCopying method in GNSS receiver component and move it to parent class

frankfu03:
The purpose of this refactoring is to make addNewByCopying() to be a generic method so that it can be used by other instrumental components (such as GNSS Antenna) when implement the update function.

### DIFF
--- a/src/client/app/shared/abstract-groups-items/abstract-group.component.ts
+++ b/src/client/app/shared/abstract-groups-items/abstract-group.component.ts
@@ -211,6 +211,30 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel>
     }
 
     /**
+     * Add a new GNSS Receiver by copying and updating values from the current one
+     */
+    addNewByCopying(event: UIEvent) {
+        if (!this.hasItems()) {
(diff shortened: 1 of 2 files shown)


src/client/app/shared/abstract-groups-items/abstract-group.component.ts
```diff
@@ -210,6 +210,30 @@ export abstract class AbstractGroupComponent<T extends AbstractViewModel>
...
+            this.copyValuesFromCurrentItem(newItemForm, currentItemForm);
+        });
+    }
+
+    /**
+     * Copy values from the current item to the new item created, and disable key fields if required.
+     *
+     * Note: cannot declare as an abstract method as not all child components will implement the Update function
+     */
+    copyValuesFromCurrentItem(newItemForm: FormGroup, currentItemForm: FormGroup): void {}
```
lbodor: Could this be declared abstract?

lbodor: I see now your answer in the comment above.

PR #708, opened 2019-06-06
Add edit function for all instrument components

frankfu03:


src/client/app/gnss-receiver/gnss-receiver-item.component.html
```diff
@@ -14,6 +14,16 @@
             (<span [ngClass]="{'hidden-xxxs': index != 0}">marked </span>to delete)
         </span>
         <span class="pull-right" [ngClass]="{'hidden-xs': gnssReceiver.dateDeleted}">
+            <button *ngIf="!isNew"
+                    type="button"
+                    [disabled]="isDeleteDisabled()"
+                    class="btn btn-xs"
+                    [ngClass]="{'btn-danger': isItemEditable, 'btn-primary': !isItemEditable}"
+                    title="Click to {{isItemEditable ? 'disable' : 'enable' }} editing of the {{getItemName()}}"
```
lbodor: It's not a simple toggle any more. It's now more like "Enable editing" and "Discard changes". (I don't think we don't need to say "Click...", because it's a button and people already know they should click it.

lbodor: I think that it would be better to have multiple buttons and just hide or show them (even though they are mutually exclusive), rather than to re-purpose a single button for different actions by changing its title and css.

src/client/app/gnss-receiver/gnss-receiver-item.component.html
```diff
@@ -14,6 +14,16 @@
...
+                    (click)="toggleItemEditFlag()">
+                <i class="glyphicon" [ngClass]="{'glyphicon-share-alt': isItemEditable, 'glyphicon-edit': !isItemEditable}"></i>
+                <span class="hidden-xs">{{getItemEditButtonName()}}</span>
+            </button>
             <button type="button" [disabled]="isDeleteDisabled()"
                   class="btn btn-danger btn-xs"
                   title="{{getRemoveOrDeletedText()}} the GNSS receiver"
```
lbodor: getRemoveOrDeletedText is perhaps confusing and inaccurate. It may be better to have separate buttons, delete and undelete, and just toggle their visibility, rather than toggle their css and title.

src/client/app/gnss-receiver/gnss-receiver-item.component.html
frankfu03: Thanks, I've made changes as suggested.

frankfu03: Not sure which way is better, but I prefer a single button with different states/actions as it uses less code and can be easily implemented in angular. If we are going to use multiple buttons, the current delete/cancel button have to be split into three buttons in a late PR: "Cancel" button for new items, "Delete" and "Undelete" buttons for existing items.

I am happy to make the changes as suggested if you like.

src/client/app/gnss-receiver/gnss-receiver-item.component.html
frankfu03: Yes, I noticed that getRemoveOrDeletedText is confusing. I thought to change (refactor) it in a separate PR. It should be getDeleteButtonName().

If we decide to use separate buttons (Cancel, Delete and Undelete), we won't need this method any more.

src/client/app/gnss-receiver/gnss-receiver-item.component.html
lbodor: OK, thanks. Why don't we leave it as is and think about it some more later.

src/client/app/gnss-receiver/gnss-receiver-item.component.html
lbodor: Yes, there are benefits of having multiple buttons. We can think about it more later.

PR #722, opened 2019-06-25
Simplify e2e test by merging 19 sitelog group testing suites into 2 suites

frankfu03:


frankfu03:
This PR is ready for review. The total e2e testing time is reduced to <10 minutes from previous 25 minutes.

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
```diff
@@ -28,7 +29,53 @@ export class GnssAntennaGroup extends LogItemGroup {
...
+    private serialNumber: string = '1616';
+    private antennaReferencePoint: string = 'BBPPAA';
+    private markerArpEastEcc: string = '1';
+    private markerArpUpEcc: string = '2';
+    private markerArpNorthEcc: string = '3';
+    private alignmentFromTrueNorth: string = '0';
+    private antennaRadomeType: string = 'SNOW_1';
+    private radomeSerialNumber: string = 'SNOW_Test';
+    private antennaCableType: string = 'SNOW_2';
+    private antennaCableLength: string = '100';
```
lbodor: Fields like these already exist in model classes. Could the model classes be re-used here?

lbodor: Is it common practice to add test data to pageobjects?

lbodor:
I think that the idea of page objects is to encapsulate information and actions about page elements. They are meant to be used across tests, so it seems wrong to include test data inside page objects.

https://github.com/angular/protractor/blob/master/docs/page-objects.md
https://medium.com/tech-tajawal/page-object-model-pom-design-pattern-f9588630800b
https://stackoverflow.com/questions/15684621/should-test-data-be-passed-from-pageobject-or-test-classes

lbodor:
I do like that you've made the tests more generic, but I suspect that we should think about putting the test driver logic into the tests, not the page objects.

src/e2e/specs/page-objects/meteorological-sensor-group.pageobject.ts
```diff
@@ -29,7 +30,62 @@ export class MeteorologicalSensorGroup extends LogItemGroup {
...
+
+        if (this.itemName !== 'Water Vapor Sensor') {
+            this.dataSamplingIntervalInput.sendKeys(this.viewModel.dataSamplingInterval.toString());
+        }
+    }
+
+    protected checkValues() {
+        TestUtils.checkInputValueEqual(this.manufacturerInput, 'Manufacturer', this.viewModel.manufacturer);
+        TestUtils.checkInputValueEqual(this.typeInput, 'Type', this.viewModel.type);
+        TestUtils.checkInputValueEqual(this.serialNumberInput, 'Serial Number', this.viewModel.serialNumber);
```
lbodor: Could string constants like 'Serial Number' be found on this.serialNumberInput objects?

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
frankfu03: Each component has a view model, but no values are defined for its fields/variables because values are used in e2e test. 

I couldn't say it is a common practice, but some people prefer to this. A page object wraps an HTML page, or fragment, and spec classes run assertion logics using methods and data defined in page objects.

src/e2e/specs/page-objects/meteorological-sensor-group.pageobject.ts
frankfu03: Yes, all string names have been wrapped into InputField objects.

frankfu03:
As suggested, the test classes have been refactored. All testing logic and data have been moved from page objects to specs testing classes.

frankfu03:
This PR is ready for review again.

src/e2e/specs/page-objects/collocation-information-group.pageobject.ts
```diff
@@ -1,20 +1,20 @@
...
         return this.itemName;
     }
+
+    public setupInputFields() {
+        this.inputFields = [
+            new InputField('instrumentationType', 'text'),
+            new InputField('status', 'text'),
+            new InputField('notes', 'textarea'),
+        ];
+    }
```
lbodor: I think that we should keep the individual fields, statusInput, notesInput, etc. This is static information, so we should represent it statically, i.e., if we make a mistake regarding some field known to us statically, we should get a compile-time error, not a runtime error.

You have a valid use-case for what your doing, however, which is to traverse these fields generically. I think we should keep the fields, but then also put them into an array so that you can get to them generically, in bulk. If putting them into an array is too much boilerplate, then you can look them up generically by some field name pattern (they may all end in Input or InputElement).

Also, it doesn't seem to me that class InputField is needed. Perhaps we can just use ElementFinders instead. Attribute "name" is probably available somewhere on the element, and "type" you only seem to use to derive "text-input" from "text" and textarea-input from "textarea".

src/e2e/specs/site-log-groups/site-log-all-components.e2e-spec.ts
```diff
@@ -0,0 +1,133 @@
...
+            new LocalEpisodicEffectGroup(),
+            new MeteorologicalSensorGroup('Humidity Sensor'),
+            new MeteorologicalSensorGroup('Pressure Sensor'),
+            new MeteorologicalSensorGroup('Temperature Sensor'),
+            new MeteorologicalSensorGroup('Water Vapor Sensor'),
+            new OtherInstrumentationGroup(),
+            new RadioInterferenceGroup(),
+            new SignalObstructionGroup(),
+            new MultipathSourceGroup(),
+        ];
```
lbodor: Should above also be part of site-log page object? It's static information with respect to the structure of the page.

src/e2e/specs/site-log-groups/test-data-mockups.ts
```diff
@@ -0,0 +1,132 @@
+import { TestUtils } from '../utils/test.utils';
```
lbodor: test-data.ts is sufficient, or mock-data.ts, test and mock is a bit redundant

src/e2e/specs/page-objects/log-item-group.pageobject.ts
```diff
@@ -56,31 +83,24 @@ export class LogItemGroup {
...
-                element(by.buttonText('Yes')).click().then(() => {
-                    console.log('Deleted ' + TestUtils.getOrdinalNumber(this.newItemIndex + 1) + ' ' + this.itemName + ' item.');
-                });
-            }
-            browser.waitForAngular();
+    public enterDateTime(dateInput: ElementFinder, datetimeValue: string) {
+        dateInput.sendKeys(datetimeValue);
+        let okButton: ElementFinder = element(by.buttonText('OK'));
+        okButton.click();
+    }
```
lbodor: Helper functions like these are useful.

src/e2e/specs/site-log-groups/site-log-all-components.e2e-spec.ts
frankfu03: OK, moved to site-log page object.

src/e2e/specs/site-log-groups/test-data-mockups.ts
frankfu03: Changed to test-data.ts.

src/e2e/specs/page-objects/collocation-information-group.pageobject.ts
frankfu03: I see your point. As suggested I have restored individual input field declaration for each page object class and removed the InputField class.

src/e2e/specs/page-objects/collocation-information-group.pageobject.ts
```diff
@@ -1,17 +1,22 @@
...
+    readonly notesInput: ElementFinder = this.newItemContainer
                     .element(by.css('textarea-input[controlName="notes"] textarea'));
 
     public constructor() {
         super('Collocation Information');
+        this.inputElements = [
+            this.instrumentationTypeInput,
+            this.statusInput,
+            this.notesInput
+        ];
```
lbodor: If it's true that every subclass must provide inputElements, then perhaps the superclass should enforce it. This can be done with via a constructor argument or via an abstract method.

src/e2e/specs/page-objects/meteorological-sensor-group.pageobject.ts
```diff
@@ -1,35 +1,59 @@
...
-    readonly accuracyHPaInput: ElementFinder = this.currentItemContainer
+    readonly accuracyHPaInput: ElementFinder = this.newItemContainer
                     .element(by.css('number-input[controlName="accuracyHPa"] input'));
-    readonly accuracyPercentRelativeHumidityInput: ElementFinder = this.currentItemContainer
-                    .element(by.css('number-input[controlName="accuracyPercentRelativeHumidity"] input'));
-    readonly accuracyDegreesCelciusInput: ElementFinder = this.currentItemContainer
+    // TODO: rename controlName "accuracyPercentRelativeHumidity" to a shorter string name as the max length of
+    // controlNname returned by ElementFinder.getAttribute('ng-reflect-name') is 30 chars, the rest are truncated
+    //readonly accuracyPercentRelativeHumidityInput: ElementFinder = this.newItemContainer
+    //                .element(by.css('number-input[controlName="accuracyPercentRelativeHumidity"] input'));
```
lbodor: Is ng-reflect-name the only option? I would expect the full control name to be available somewhere. Doesn't
```
readonly accuracyPercentRelativeHumidityInput: ElementFinder = this.newItemContainer..element(by.css('number-input[controlName="accuracyPercentRelativeHumidity"] input'));
```
work? If it does, we can leave it in, and just exclude the field from inputFields.

src/e2e/specs/site-log-groups/responsible-parties.e2e-spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    it('expect should be able to add new items or backup values for existing items', () => {
+        console.log('a) Add new items or backup values for existing items');
+        siteLogPage.siteInformationHeader.click().then(() => {
+            siteLogPage.responsibleParties.map((responsibleParty: ResponsiblePartyGroup) => {
+                responsibleParty.groupHeader.click().then(() => {
+                    if (responsibleParty.canAddNewItem) {
+                        responsibleParty.items.count().then((value: number) => {
+                            responsibleParty.noOfItems = value;
+                            console.log('    Number of ' + responsibleParty.itemName
```
lbodor: Nitpick: four spaces is generally not half a tab, because a tab character can be configured to be rendered as any number of spaces by the user.

src/e2e/specs/page-objects/collocation-information-group.pageobject.ts
frankfu03: Yes, each subclass must provide inputElements. I did what you mentioned above. I declared an abstract method setInputElements() and invoke it in the constructor of the superclass. Each subclass only needs to implement the abstract method. But it turnout that the inputElements is null in subclasses. It would be working fine in Java, but not in typescript.

lbodor: Try getInputElements(). I'm not sure what went wrong when you did it earlier, but generally abstract getters and better than abstract setters, because you know that a getter will do the right thing (it's type signature says it must return a list of input fields), but when calling a setter, you are relying on a side-effect (the implementation of the setter modifying a field), whose outcome you can't enforce.

src/e2e/specs/page-objects/meteorological-sensor-group.pageobject.ts
frankfu03: Yes, it seems that ng-reflect-name the only option to get the controlName o fthe input. Both getAttribute('controlName') and getAttribute('control-name') return null.

accuracyPercentRelativeHumidityInput does not work as its controlName "accuracyPercentRelativeHumidity" is 31 chars long. testing indicate that getAttribute('bg-reflect-name') only return 30 chars: "accuracyPercentRelativeHumidit". If we want to test this input filed, I can fix it in a separate PR.

frankfu03: I found "ng-reflect-name" from getOuterHtml(). For example,

`<input _ngcontent-c4="" class="form-control ng-untouched ng-pristine ng-valid" type="text" ng-reflect-maxlength="10000" ng-reflect-name="receiverType" maxlength="10000">`

src/e2e/specs/site-log-groups/responsible-parties.e2e-spec.ts
frankfu03: White spaces are replace by "\t" or "\t\t" in console.log().

lbodor: :+1:

src/e2e/specs/page-objects/meteorological-sensor-group.pageobject.ts
frankfu03: After changed to find parent elements of input tags, we can use getAttribute('controlName') to get controlName values now.

lbodor: Great result.

src/e2e/specs/page-objects/collocation-information-group.pageobject.ts
frankfu03: Renamed the abstract method to setInputElements(). This method can not be called in parent class (a list of nulls), but can be invoked in subclasses.

PR #723, opened 2019-07-09
E2e update tests

frankfu03:


src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
```diff
@@ -30,23 +30,29 @@ export class GnssAntennaGroup extends LogItemGroup {
 
     public constructor() {
         super('GNSS Antenna');
-        this.getInputElements();
     }
 
     public getInputElements() {
+        if (this.noOfItems > 0) {
+            this.canAddNewByUpdate = true;
```
lbodor: Would this be same as "canUpdate"?

src/e2e/specs/site-log-groups/site-log-all-components.e2e-spec.ts
```diff
@@ -42,12 +42,23 @@ describe('All Site Log Components', () => {
...
                     });
-                });
+                } else {
+                    siteLogGroup.addNewItemButton.click().then(() => {
+                        console.log('\tAdd a new ' + siteLogGroup.itemName + ' item');
+                        siteLogGroup.inputElements.map((inputElement: ElementFinder) => {
+                            TestUtils.changeInputValue(inputElement, testData);
+                        });
+                    });
+                }
```
lbodor: Do we still need a test to "add new" (without update) for antennas and receivers?

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
frankfu03: Ok, renamed "canAddNewByUpdate" to "canUpdate".

src/e2e/specs/site-log-groups/site-log-all-components.e2e-spec.ts
frankfu03: For antenna or receiver, we run conditional e2e testing: update if there is a current item available; addNew if no any item available. In most cases, it will run update.

frankfu03: If we are sure that there are always some antenna and receiver items available in e2e test, we can simplify the e2e tests for antenna and receiver by removing addNew tests.

lbodor: If we only run update tests on antenna and receivers, won't we then miss out on testing all the input fields?

frankfu03: Yes, 2 fields in receiver and 5 fields in antenna won't be tested in update as these fields are disabled.

lbodor: From that point of view, if only testing one case, it would seem more efficient to only test addNew instead of only update.

frankfu03: Ok, then I'll close this PR.

lbodor: Why don't we test both cases, update and addNew?

frankfu03: Testing of update of receiver/antenna can be done in another spec testing class, eg., site-log-update-items.e2e-spec.ts.

frankfu03: How about I rename the testing class from "site-log-all-component.e2e-spec.ts" to "site-log-add-new-items.e2e-spec.ts"?

site-log-groups/
                responsible-parties.e2e-spec.ts
               site-log-add-new-items.e2e-spec.ts
                site-log-update-items.e2e-spec.ts

lbodor: Yes, I think that's a good idea.

frankfu03:
e2e test of update receiver/antenna are now running in a new e2e-spec testing class: site-log-update-items.e2e-spec.ts. This PR is ready for review.

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
```diff
@@ -49,4 +48,20 @@ export class GnssAntennaGroup extends LogItemGroup {
             this.notesInput,
         ];
```
lbodor: There is duplication here. I think that getInputElements should call getUpdateElements, because it is not a coincidence that all updateElements are also inputElements.

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
```diff
@@ -49,4 +48,20 @@ export class GnssAntennaGroup extends LogItemGroup {
...
+        this.inputElements = [
+            this.markerArpEastEccInput,
+            this.markerArpUpEccInput,
+            this.markerArpNorthEccInput,
+            this.alignmentFromTrueNorthInput,
+            this.antennaCableTypeInput,
+            this.antennaCableLengthInput,
+            this.notesInput,
+        ];
+    }
```
lbodor: Getters should return values, preferably without modifying object state.

frankfu03: we may change to addInputElements() and addUpdateElements(), or setInputElements() and setUpdateElements(). Which one do you prefer or any other suggestions?

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
frankfu03: Do you mean to change this.inputElements to this.updateElements in getUpdateElements()?

Variable this.inputElements was reused for both methods: getInputElements() should call getUpdateElements().

frankfu03: Maybe I can change getInputElements() to getAddNewElements(). Any thoughts?

lbodor: Let's consider removing inputElements variable. The superclass declares two abstract methods, getInputElements() and getUpdateElemets(), forcing the subclasses to provide both lists. Then the subclasses provide the implementations like this (pseudocode):

```
And who ever needs input elements or update elements just calls the methods, no variable. Would that work
getUpdateElements(): [ElementFinder] {
    return [a, b];
}

getInputElements(): [ElementFinder] {
    return this.getUpdateElements() ++ [c, d];
}
```

lbodor: I think we should called them input fields rather than elements. I know the type is ElementFinder, but that's generic because it be used to find any page element, but we are looking only for input fields. Some examples:
getAllInputFields(), getUpdatableInputFields
getInputFields, getUpdateableInputFields

frankfu03: Ok, I'll change variable names from inputeElements to inputFields, and change method names from getInputElements() to getAllInputFields(), from getUpdateElements() to getUpdateableInputFields().

As "Getters should return values" as metioned in below review, may I change "get" to "set" or "add"?

lbodor: My suggestion is to remove the variables and to implement getters, no variables, no setters, and no getters that set.

lbodor: Would that work? Abstract setters don't really make much sense. If you want to require subclasses to provide some values, then you can either force them to implement a getter (getters provide values) or you can force them to provide these values by declaring a constructor in the superclass that would require these values to be passed in.

lbodor: If the superclass declares an abstract setter, a subclass must implement it, but there is no way to guarantee that the setter in the subclass does the right thing.

frankfu03: I used setter and getter for input fields, but failed in responsibleParty groups. Take "site contacts" as an example, when adding a new item, the new one is the first item. After saving and reloading, the "new" item could be anywhere, from 1st to last item. so get input fields will return the old fields.

lbodor: To me that looks like a separate problem. Input field element finders in responsible party group object capture the input fields of the first contact. Saving the input fields in the page object, as is happening currently, doesn't solve the general problem, it just solves the problem in the case of the one particular e2e test. If the work-around is specific to the test, then could we save the input fields in the test itself, not the page objects?

Either way, the general problem of how to order multiple contacts remains.

frankfu03: I think you are right. So, I am going to create another PR to do the changes: remove variable inputElements and change method getUpdateElements() to return ElementFinder[], etc.

lbodor: :+1:

src/e2e/specs/page-objects/gnss-antenna-group.pageobject.ts
frankfu03: Variable "canUpdate" has been removed after rebasing against previous PR (refactor-e2e-input-fields).

PR #724, opened 2019-07-14
Remove variable inputElements and change getInputElements() to return…

frankfu03:
… ElementFinder array

### DIFF
--- a/src/e2e/specs/page-objects/responsible-party-group.pageobject.ts
+++ b/src/e2e/specs/page-objects/responsible-party-group.pageobject.ts
@@ -33,8 +33,8 @@ export class ResponsiblePartyGroup extends LogItemGroup {
         super(partyName);
         this.hasEndDateInput = false;
         this.canAddNewItem = false;
-        this.getInputElements();
         this.backupModel = {};
+        this.updateInputFields();
 
         this.items = element(by.cssContainingText('.panel-level-2', this.getGroupName()))
                             .all(by.css('gnss-responsible-party-item'));
@@ -67,7 +67,7 @@ export class ResponsiblePartyGroup extends LogItemGroup {
                 element.getAttribute('value').then((value: string) => {
                     if (value === positionNameValue) {
                         this.newItemIndex = index;
-                        this.getInputElements();
+                        this.updateInputFields();
                         console.log('\tNote: the index for the new ' + this.getGroupName()
                                     + ' after saving/reloading is: ' + index);
                     }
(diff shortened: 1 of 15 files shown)


src/e2e/specs/page-objects/responsible-party-group.pageobject.ts
```diff
@@ -76,7 +76,27 @@ export class ResponsiblePartyGroup extends LogItemGroup {
...
+            this.primaryPhoneInput,
+            this.secondaryPhoneInput,
+            this.faxInput,
+        ];
+    }
+
+    /**
+     * Update all input fields as new item may change its position after saving
+     */
+    private updateInputFields(): void {
```
lbodor: Would it be possible to eliminate this method "update" method as well, if we were to replace:
```
public updateNewItemElements(positionNameValue: string): void {
```
with
```
public getAllInputFields(positionNameValue: string): ElementFinder[] {
```
?

lbodor: The responsible party group object would then be different from others only because it can can give you input fields for any position, rather than just the top. It would become quite a small difference, and arguable other groups could one day give you input elements for any position too, eliminating the special treatment of responsible parties. Then it's up to the tests to ask always for the top position or for whatever position they want.

lbodor: Instead of setting "top" input fields to some arbitrary position and then fetching the top position, you would just fetch input elements from any arbitrary position.

frankfu03: I tried to merge updateNewItemElements(positionNameValue: string) into getAllInputFields(positionNameValue: string), but failed because the update of input fields is actually done inside element.getAttribute('value') which returns a Promise. It seems that we couldn't return a list of input fields after the promise (return statement will run before the promise completes), and it is complicate to return input fields inside the promise (eg., return a Promise wrapped with input fields).

So we may leave it as is now and solve it later if having new ideas.

lbodor: Yes, let's leave it. Thanks for trying. I will look again at some stage at ordering multiple contacts, which could also simplify the testing logic.

frankfu03: Need approval before merging.

lbodor: Sorry I didn't realise I hadn't approved. Thanks for prompting.

PR #725, opened 2019-07-15
 Add e2e tests on site-administration component

frankfu03:


src/e2e/specs/site-log-groups/site-log-all-components.e2e-spec.ts
```diff
@@ -25,7 +25,7 @@ describe('All Site Log Components', () => {
 
         browser.get(selectSitePage.url);
         browser.waitForAngular();
-        loginActions.login('user.a', 'gumby123A');
+        loginActions.loginAs('user.a', 'gumby123A');
```
lbodor: I think "login" is better and a more traditional name for this method. Is there also "login", which doesn't take any arguments?

frankfu03: No, there is no login() method that does not take any argument.

`login(username, password)`: skip login if already done, otherwise sign in with provided credential;
`loginAs(username, password)`: logout first if already signed in as a different user, and then sign in with provided credential.

We need to sign in as user.a for responsible parties and site log groups, but as user.x for site admin. So, we need logout and then login.

How about change `loginAs()` to `loginForce()`?

frankfu03: In this PR, the e2e tests are running in the order: (1) responsible-parties.e2e-spec.ts -> (2) site-administrations.e2e-spec.ts -> (3) site-log-all-components.e2e-spec.ts. The (1) and (3) are logged in as user.a while (2) as user.x. Previously, there is no problem without (2) using login(). After adding (2) in this PR, transit from (1) to (2) and then (3) need logout and then login. 

 `loginAs('user.a', 'gumby123A')` can do both logout and login. Otherwise, e2e test will fail.

lbodor: Right. A better way might to either add logout() (which would one day be useful to test logging out) and call it before calling login if you wish to change the user. Or modify login function to first check who is logged in and do nothing if it is being asked to log in as the same user or logout and log in as a different user if required.

The driver for this comment is not to introduce new names that don't have an obvious difference in meaning. You and I can decide what "loginAs" means and how it is different from "login", but it won't be obvious compared to just having "login" and "logout", or just "login" that does the right thing.

frankfu03: Thanks. I changed back to login() and added logout() to afterAll() for each spec class.

login() and loginAs() are defined in e2e/specs/utils/login.actions.ts. If we need to do any changes to them, we can do it in a separate PR.

lbodor: OK, thanks.

PR #727, opened 2019-07-17
E2e misc issues

frankfu03:


frankfu03:
The name "src/e2e/specs/page-objects/**log-item-group**.pageobject.ts" is not quite meaningful. It should be rename to "abstract-group", similar to what we defined in "src/client/app/shared/abstract-groups-items/**abstract-group**.component.ts". What do you think?

PR #732, opened 2019-10-01
User defined receiver, antenna and radome type

dotzhou:


src/client/app/gnss-antenna/gnss-antenna-item.component.ts
```diff
@@ -19,6 +19,7 @@ export class GnssAntennaItemComponent extends AbstractItemComponent {
      * The GNSS Antenna in question.
      */
     @Input() antenna: GnssAntennaViewModel;
+    pattern: string = '^[A-Z0-9._-]+( [A-Z0-9._-]+)* +[A-Z]{4}$';
```
umeat: RADOME should probably be allowed to have numbers.

dotzhou: Indeed, I change

PR #734, opened 2019-10-03
Use prefetched codelists for antenna & receiver typeahead inputs

frankfu03:


frankfu03:
This PR should be merged before typeahead-form-input (Create a generic typeahead-input to replace antenna-type-input and receiver-type-input components #762).
Currently antenna-radome and receiver codelists are hijacked from src/client/app/shared/geodesyml-codelist/geodesyml-codelist.service.ts. This PR is to pass both codelists from site component to antenna / receiver group/item compoments.

lbodor:
Yes, I agree. thanks for pointing that out.

PR #737, opened 2019-12-01
Deploy to dev on merging into master branch

frankfu03:


frankfu03:
Yes, it will. So we have to deploy the upgrade of geodesyml 0.5 in gws first. I already submitted another PR for this.

frankfu03:
No, it won't as both site manager and gws are currently using geodesyml v.0.4. So this PR should be merged before another PR (update-geodesyml-to-v0.5).

PR #740, opened 2020-01-23
Add associated document http service for uploading/deleting documents…

frankfu03:
… to/from storage

### DIFF
--- a/src/client/app/shared/associated-document/associated-document.service.spec.ts
+++ b/src/client/app/shared/associated-document/associated-document.service.spec.ts
@@ -1,0 +1,72 @@
+import { ReflectiveInjector } from '@angular/core';
+import { BaseRequestOptions, ConnectionBackend, Http, Response, ResponseOptions } from '@angular/http';
+import { MockBackend } from '@angular/http/testing';
+import { User } from 'oidc-client';
+
+import { AssociatedDocumentService } from './associated-document.service';
+import { ConstantsService } from '../global/constants.service';
+import { UserAuthService } from '../global/user-auth.service';
+
+export function main() {
+    describe('Associated Document Service', () => {
+        let service: AssociatedDocumentService;
+        const statusCreated = 201;
+        const statusNoContent = 204;
+        const documentName = 'AssociatedDocument_Upload_Testing.txt';
+        const mockUserService = new (class MockUserService {
+            userSettings: any = {id_token: '', session_state: '', access_token: '',
+                                 refresh_token: '', token_type: '', scope: '',
+                                 profile: null, expires_at: 0, state: {}};
+            user: User = <User>this.userSettings;
+        })();
+
+        beforeAll(() => {
+            let injector = ReflectiveInjector.resolveAndCreate([
+                AssociatedDocumentService,
+                ConstantsService,
+                UserAuthService,
+                BaseRequestOptions,
+                MockBackend,
+                {
+                    provide: Http,
+                    useFactory: function(backend: ConnectionBackend, defaultOptions: BaseRequestOptions) {
+                        return new Http(backend, defaultOptions);
+                    },
+                    deps: [MockBackend, BaseRequestOptions]
+                },
+                {
+                    provide: UserAuthService, useValue: mockUserService
+                },
+            ]);
+            service = injector.get(AssociatedDocumentService);
+
+            let backend = injector.get(MockBackend);
+            let responses: Response[] = [
+                new Response(new ResponseOptions({
+                    body: 'document-url:' + documentName,
+                    status: statusCreated
+                })),
+                new Response(new ResponseOptions({status: statusNoContent}))
+            ];
+            backend.connections.subscribe((connection: any) => {
+                let response = responses.shift();
+                connection.mockRespond(response);
+            });
+        });
+
+        it('should upload a document to data storage', () => {
+            let file = new File(['A dummy file for testing.'], documentName, {type: 'text/plain'});
+            service.uploadDocument(documentName, file).subscribe((response: Response) => {
+                const fileReference = response.text();
+                expect(response.status).toBe(statusCreated);
+                expect(fileReference).toContain(documentName);
+            });
+        });
+
+        it('should delete the uploaded document from data storage', () => {
+            service.deleteDocument(documentName).subscribe((response: Response) => {
+                expect(response.status).toBe(statusNoContent);
+            });
+        });
+    });
+}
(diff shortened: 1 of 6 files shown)


frankfu03:
This PR is ready for review.

src/client/app/shared/associated-document/associated-document.service.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    provide: UserAuthService, useValue: mockUserService
+                },
+            ]);
+            service = injector.get(AssociatedDocumentService);
+
+            let backend = injector.get(MockBackend);
+            let responses: Response[] = [
+                new Response(new ResponseOptions({
+                    body: 'http://localhost:4572/gnss-metadata-document-storage-local/' + documentName
```
lbodor, line 45: This URL looks real, but in this context it isn't. There is no localhost:4572 or gnss-metadata-document-storage-local in this context. Could we use a string like "document-url", just so nobody gets confused. Also, what about returning the url in the "location" response header? It would be a change to the API as well, but I think that that's the purpose of the location response header.

src/client/app/shared/associated-document/associated-document.service.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                })),
+                new Response(new ResponseOptions({status: statusOK}))
+            ];
+            backend.connections.subscribe((connection: any) => {
+                let response = responses.shift();
+                connection.mockRespond(response);
+            });
+        });
+
+        it('should upload a document to S3 bucket', () => {
```
lbodor, line 55: We shouldn't not be mentioning s3 buckets, it's an implementation detail.

src/client/app/shared/associated-document/associated-document.service.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+            let file = new File(['A dummy file for testing.'], documentName, {type: 'text/plain'});
+            service.uploadDocument(documentName, file).subscribe((response: Response) => {
+                const fileReference = response.text();
+                expect(fileReference).toContain(documentName);
+            });
+        });
+
+        it('should delete the uploaded document from S3 bucket', () => {
```
lbodor, line 63: There is no bucket.

src/client/app/shared/associated-document/associated-document.service.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Headers, Http, Response } from '@angular/http';
+import { Observable } from 'rxjs/Rx';
+import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
+import { User } from 'oidc-client';
+
+import { ConstantsService } from '../global/constants.service';
+import { UserAuthService } from '../global/user-auth.service';
+
+/**
+ * This class provides a http service for uploading/deleting associated documents to/from AWS S3 bucket.
```
lbodor, line 11: No bucket.

src/client/app/shared/associated-document/associated-document.service.spec.ts
frankfu03: Ok, I'll change "s3 bucket" to "storage".

src/client/app/shared/associated-document/associated-document.service.spec.ts
frankfu03: Good, will do the change same as above.

src/client/app/shared/associated-document/associated-document.service.spec.ts
frankfu03: Yes, we can use any string as the url because only the document name is checked in the test.
Not sure what do you mean for the "location" response header. The ResponseOptions class in Angular does not have the "location" field. See the [angular doc](https://v2.angular.io/docs/ts/latest/api/http/index/ResponseOptions-class.html).

lbodor: https://en.wikipedia.org/wiki/HTTP_location
https://restfulapi.net/http-status-201-created/

frankfu03: Thanks. It would be easy to use a string to represent the url for the mock response.

src/client/app/shared/associated-document/associated-document.service.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    provide: UserAuthService, useValue: mockUserService
+                },
+            ]);
+            service = injector.get(AssociatedDocumentService);
+
+            let backend = injector.get(MockBackend);
+            let responses: Response[] = [
+                new Response(new ResponseOptions({
+                    body: 'document-url:' + documentName,
```
lbodor, line 46: Why not "location" response header? Isn't that the standard way of delivering this type of information?

frankfu03: I tried to change "body" to "location" but got an error indicating no field "location" in ResponseOptions class.

src/client/app/shared/associated-document/associated-document.service.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { AssociatedDocumentService } from './associated-document.service';
+import { ConstantsService } from '../global/constants.service';
+import { UserAuthService } from '../global/user-auth.service';
+
+export function main() {
+    describe('Associated Document Service', () => {
+        let service: AssociatedDocumentService;
+        const statusCreated = 201;
+        const statusNoContent = 204;
+        const documentName = 'AssociatedDocument_Upload_Testing.txt';
```
lbodor, line 15: This is also a bit misleading. The returned value is a URL, so if anything it should look like a URL, rather than a file name. It should look like nothing ("document-url") or it should look like a URL ("https://path/to/document.txt").

src/client/app/shared/associated-document/associated-document.service.spec.ts
lbodor: That's because "location" is a header name. You should find out how to set response headers. There will be a corresponding change to gws-core.

src/client/app/shared/associated-document/associated-document.service.spec.ts
frankfu03: Yes, the returned value is a URL to the document in the S3 bucket. In the test we only check the document name and ignored the URL path. Before it was "http://localhost:4572/gnss-metadata-storage-local/AssociatedDocument_Upload_Testing.txt", which is the actual value to be returned if we do not use mock response. Maybe I misunderstood your comments.

src/client/app/shared/associated-document/associated-document.service.spec.ts
frankfu03: OK, I'll do an investigation on it.

frankfu03: I have changed to use "location" response header in response to the changes made to gws-core.

PR #742, opened 2020-04-08
Add site image component

frankfu03:


src/client/app/shared/thumbnail-image/full-size-image.component.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.images = [];
+        this.currentImage = null;
+    }
+
+    close() {
+        this.closeEvent.emit();
+    }
+
+    showPrevImage() {
+        this.currentImageIndex --;
```
lbodor, line 53: no space before `--`

src/client/app/shared/thumbnail-image/full-size-image.component.ts
```diff
@@ -0,0 +1,67 @@
...
+        if (this.currentImageIndex < 0) {
+            this.currentImageIndex = this.images.length - 1;
+        }
+        this.currentImage = this.images[this.currentImageIndex];
+    }
+
+    showNextImage() {
+        this.currentImageIndex ++;
```
lbodor, line 61: no space

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
```diff
@@ -0,0 +1,130 @@
...
+    styleUrls: ['thumbnail-image.component.css'],
+})
+export class ThumbnailImageComponent implements OnChanges, OnInit, OnDestroy {
+
+    thumbImageWidth: number = 200;
+    thumbImageHeight: number = 150;
+    imagePanelHeight: number = 320;
+
+    @Input() set imageSize(data: any) {
+        if (data && typeof (data) === 'object') {
```
lbodor, line 31: no parenthesis: `typeof data`

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
```diff
@@ -0,0 +1,130 @@
...
+            if (data.hasOwnProperty('width') && typeof (data['width']) === 'number') {
```
lbodor, line 32: https://dev.to/andreasbergqvist/typescript-get-types-from-data-using-typeof-4b9c

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    setSliderWidth() {
+        this.imagePanelHeight = 2 * this.thumbImageHeight + 20;
+    }
+
+    deleteImage(imageIndex: number) {
+        let image: ImageObject = this.images[imageIndex];
+        let msgHtml: string = '<div><div class="title">Confirm Site Image Deletion</div>'
+            + '<div class="body"><p/><p>"' + image.title + ' ' + image.createdDate
+            + '" will be permanently deleted from system storage and database.'
```
lbodor, line 99: I think that `system storage and database` is too detailed for the user. `Permanently deleted` would be enough. We don't need to tell the user that there is system storage and that there is a database.

src/client/app/site-image/site-image.component.html
```diff
@@ -0,0 +1,225 @@
...
+                        <tr role="row" *ngFor="let image of siteImagesDeleted; let i = index">
+                            <td name="description">{{image.description}}</td>
+                            <td name="originalName">{{image.originalName}}</td>
+                            <td name="fileReference">{{image.fileReference}}</td>
+                            <td name="createdDate">{{miscUtils.formatDateTimeSimple(image.createdDate)}}</td>
+                        </tr>
+                        </tbody>
+                        <tfoot>
+                        <tr>
+                            <td colspan="4" class="text-warning">Note: you must click "Save" button to save changes to sitelog geodesyML</td>
```
lbodor, line 74: Having to click "Save" to save changes is not specific to images, so I think that the message shouldn't be so closely associated with the images UI. If we want to inform the user that changes aren't actually saved until "Save" is clicked, then we should do it somewhere else (in the header, in the margin, top of page), because the message applies equally to images as to receivers, site location, and everything else on the page.

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
frankfu03: Yeah, it is a good way to get the type of a object.
Not sure how to make the changes here. "data" is an input passed from its calling component. It is expected that "data" is a json object containing two number attributes: width and height of a thumbnail image. 
Instead of a json object, maybe we can change to have two @Input variables: imageWidth and imageHeight?

lbodor: I just sent the link as an example of how "typeof" doesn't need parenthesis. It's more like an operator than a function call.

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
frankfu03: Good. `system storage and database` is removed from the message string.

lbodor:
I'm not sure that including link into the s3 bucket is appropriate given that we send out site logs to third-parties. Renaming the bucket would invalidate all links. I'm not sure what the best solution would be. Perhaps we should wrap hide those links behind a get request to GWS, they would look better as well. But then again, the GWS endpoint will also change at some point.

Perhaps when the image links change, we need to regenerate all the site logs, so that third-parties get updated version when they poll for changes.

lbodor:
image

image

I'm not sure that the subheading should be changing from "Upload site images" to "Add site image links". I think that the top-level action is something like "New image", with there being two methods to add a new image, by uploading a file or by providing a link to a file. In both cases, the user is still just doing one thing, which is adding a new site image.

I'm also a bit concerned that the UI for images is different than the UI for managing receivers and antennas, etc. For example sections "Current Site Images" and "Site Images Uploaded To System Storage", do have their equivalent representation when editing antennas and receivers, they are just  combined into one view. And following the already established pattern would give us a "New" button on the right-hand side of "Current Images".

frankfu03:
As for the including links into the s3 bucket, it would be good if we can wrap hide them behind a get request to GWS. But we need a html img link to display images on web directly, not sure if a get request to GWS will do so. I'll have a try.

frankfu03:
As for the layout of Site Image component, maybe we can change it to:
```
 Site Images                                   New _(button)_
    Current Site Images
    Previous Site Images
```
Initially the "Upload Site Images" panel is hidden. Clicking on the "New" button will show this panel. When users saved the sitelog, this panel will be hidden again.

src/client/app/shared/thumbnail-image/thumbnail-image.component.ts
frankfu03: Yeah, all parenthesis have been removed. I thought you refer to the "any" data type.

src/client/app/site-image/site-image.component.html
frankfu03: Good, I'll remove the warning text.

But I think that uploading a site image is different from modifying other parts of the sitelog. If users upload a image into the s3 bucket but forget to save the sitelog, the image uplaoded will become orphan.

lbodor:
Yes, I think that placing a "New" button to the right of "Site Images" would be make the UI more consistent.

src/client/app/site-image/site-image.component.html
lbodor: That's true, but I think that the orphans arise from our choice of implementation, and we don't necessarily need to worry the user being careful not to make any orphans. We should just be ready to delete them periodically.

lbodor:
Could we combine the list of pending images and the list of current images into one? This is how receivers behave. The user can click on "New" receiver twice and get two pending receiver records shown in the same list.

When adding images, the new images could maybe appear right along side current images, but outlined in green to denote that they are new, like new receivers are green. This would remove the need for the table of pending images. All other items in on the page combine the pending and the existing items into a single list.

lbodor:
How can a user delete an image?

frankfu03:
Users can delete historic images and can also delete pending images. Current images are not allowed to be deleted (we can change this rule if needed).

frankfu03:
It is a good idea to remove the pending images list and combine them with others, marking the new ones with green color.
Images are classified by their date_created. New imaged may not be the latest. Only the latest image of a type will be in "Current Images" panel, all other older images of the same type are in "Historic Images" panel. 
So, if we combine pending images with others, the new images could be shown in the "Current Images" panel if their date_created is the latest (as a result, some current images of the same types will be pushed down to the "Historic Images" panel. And some of the new images could be shown in the "Historic Images" panel if their date_created are not the latest.

lbodor:
Exactly.

umeat:
I think we should be able to delete current images, in much the same way as we can delete current receivers, if possible.

frankfu03:
OK. So both current and previous images can be deleted by authorized users.

jmettes:
I agree the upload UI should be consistent with the other components via a "New" button. Also it shouldn't possible to do the upload process if the user is not logged in, currently I can click the Upload panel and expand it without logging in. The Add and Update buttons for receivers and antennas are still present, although greyed-out & not clickable. Perhaps this could be the same for the images.

Are there currently any Pending images? I don't see any. I wonder if it would be useful for pending images to be greyed-out a little to show they are not active.

I agree the changing header is a bit jarring when toggling between "add new images" / "site links". The only thing that changes is "Select image" picker, vs "Image link" text box. Perhaps they can both be available below each other. I feel like I've seen that kind of thing where you can either click a file picker or enter a URL. I think "Image URL" would be more clear than "Link" - I was initially confused whether this was a special thing that gets linked to some metadata or something.

The historic West Facing image for ALIC looks broken.

jmettes:
Does it make sense to have an "Add" button, as well as an "Update" button. At first glance, it wasn't clear to me how the UI knew about whether my image was Front Facing or whatever.

From my understanding, there are essentially a number of prescribed views that each site **should** have (is this true? or only for the ones we own?). They are in essence "slots" for photos that should be filled in, and is kind of poor form to not have them there. Perhaps it makes sense to arrange the images in such a way that it makes clear this formality. For example showing the North Facing image at the top, south on the bottom, West on the left, etc. Where there is no image uploaded, perhaps it still makes sense to present their empty slot to indicate this is missing.

For example, I've noticed this convention with some other GNSS network sites: https://www.ngs.noaa.gov/cgi-cors/corsage.prl?site=MOAL

Just my opinion - I do admin it does look a little clunky to arrange them like that, but it does have something going for it. Are the tiles always ordered in the same way? I think this would help a user to quickly know spot the image they are after for comparisons.

frankfu03:
Good points, Jonathan. 

- To mark a pending image, a "New" tag will be put on it. I can also gray-out pending images if needed.
- Only the user who uploaded a image but not yet saved can see the pending image. To others, it is orphan image.
- `The historic West Facing image for ALIC looks broken`: it was deleted but not saved. I did this for testing. I try to show it properly so that we can still remove it from DB.
- I'll change "Image Link" to "Image URL" as suggested.
- Happy to hear more on Image Description, like "North Facing", showing images in order, etc.

frankfu03:
The function of deleting images from s3 has been removed from the code. It seems working well.

umeat:
If you set the Image Description - then change from Image File to Image URL (or vice versa) it removes your selection in Image Description. That shouldn't happen.

PR #751, opened 2020-05-05
Notice board component

frankfu03:


frankfu03:
The new component notice-board can be reused and will be merged into master branch too. By default it is empty and thus won't be shown. If we later want to publish a notice, just add a notice title and a notice text to the component.

PR #764, opened 2020-07-15
Change ng2-bootstrap v.1.3.3 to ngx-bootstrap v2.0.5

frankfu03:


frankfu03:
ngx-bootstrap's typeahead has a reported issue: [https://github.com/valor-software/ngx-bootstrap/issues/4350](https://github.com/valor-software/ngx-bootstrap/issues/4350) - Typeahead highlight displays <strong> tag when empty spaces added at the end of query and does not escape html tags #4350.
It appears that typeahead works fine when there is only one white space between words in dropdown menu. In our cases of antenna-radome types, there are sometimes two or more consecutive white spaces between antenna and radome types. 
Ted forked ngx-bootstrap v.2.0.5 to ngx-boostrap-ga and provided a temporary solution to this issue by removing <strong> and </strong> html tags from the dropdown menu. As a result, it disables the feature of highlighting search text matched in the dropdown menu. So this is a drawback of his solution.
There are two options: 1) still use the typeahead from the forked ngx-bootstrao-ga; 2) use the typeahead of ngx-bootstrap v2.0.5: remove extra white spaces from antenna-radome types and reformat antenna-radome types when saving sitelog.

lbodor:
Could we also fix it in a way that it all works even when there are multiple spaces between the words?

frankfu03:
Do you mean that keep the forked ngx-bootstrap-ga and modify the code to restore the highlight search text feature?

lbodor:
I guess. My order of preference is:

1. have it all work, fork or no fork
2. keep multiple spaces, but ditch highlighting, because we can't do or don't want to spend the time on 1)
3. restore highlighting, but ditch multiple spaces

I know that the maintaining the fork is not ideal, but I suspect that fiddling with the spaces can be error prone.

lbodor:
I think that fiddling with the spaces, unless we record somewhere how many we removed and restoring that many, will amount to implementing some kind of judgement around what is a correct value to put into a sitelog, which I'm not sure we shouldn't do. The correct values come from the third-party lists we parse, and if they include a bad value, I think we shouldn't fix that, but expose it.

frankfu03:
OK. I'll use the forked ngx-bootstrap-ga.

lbodor:
ok

PR #768, opened 2020-08-27
Remove auto scroll function and its library scroll-into-view

frankfu03:
I have deployed this PR to DEV. You may test to see the changes via: 
[GNSS-Site-Manager](https://dev.gnss-site-manager.geodesy.ga.gov.au)


lbodor:
This change looks good to me, but it would be good to get some feedback from ops and infrastructure.

frankfu03:
Yes, we'll keep it for a couple of weeks until we get some feedback.

umeat (approved):
This seems so much nicer to me.

PR #772, opened 2020-09-16
Merge date-input component into datetime-input component

frankfu03:


src/client/app/shared/form-input/datetime-input.component.ts
```diff
@@ -46,13 +54,15 @@ export class DatetimeInputComponent extends AbstractInput implements OnInit {
...
+        validators.push(new DatetimeFormatValidator(this.showTime));
+        if (this.showTime) {
+            if (this.controlName === 'endDate') {
+                let startDateControl: FormControl = <FormControl>this.form.controls.startDate;
+                validators.push(new DatetimeRangeValidator(this.dateType, startDateControl, true));
+            } else if (this.controlName === 'startDate') {
+                let endDateControl: FormControl = <FormControl>this.form.controls.endDate;
+                validators.push(new DatetimeRangeValidator(this.dateType, endDateControl, false));
+            }
         }
```
lbodor, line 66: Is some equivalent validation not needed when `showTime` is false?

src/client/app/shared/form-input/datetime-input.component.ts
```diff
@@ -19,12 +19,14 @@ export class DatetimeInputComponent extends AbstractInput implements OnInit {
...
 
+    @Input()
+    showTime: boolean = true;
+
     public datetimeModel: Date;
     public hours: number = 0;
     public minutes: number = 0;
     public seconds: number = 0;
-    public datetimeLength: number = 19;
-    public dtPickerHeight: number = 380;
```
lbodor, line 27: :+1:

src/client/app/shared/form-input/datetime-input.component.ts
frankfu03: yes, site image's created date (Date Taken in UI) does not include time.

lbodor: It looks like DatetimeRangeValidator checks whether startDate and endDate are the the right way around, which generally could be relevant even if the datetime values don't include times of day. It looks like here you know that if showTime is false, then there isn't a startDate and an endDate, because showTime happens to be true only for a UI component that is a single value, not a range.

It seems that the implementation of datetime-input-component is reaching out to beyond what it's mean to be aware of, which is how it is used. If this were in a library, then this component couldn't assume that showTime was true for single values and that startDate and endDate represented a range.

It looks like we may be missing a generic datetime-ranage-component. Maybe something to think about at another time -- or not, writing generic UI components is hard.

lbodor (approved):
It's much better that we have one component now, rather than two.

src/client/app/shared/form-input/datetime-input.component.ts
frankfu03: Good points. Maybe it is good to keep datetime and date input components separate, not to merge them.

lbodor: I think it's good to merge them, but I think that we are missing a datetime-range component that can includes two datetime components. Datetime format validation would be on the datetime component, but range validation should be on the datetime-range component, and should work for both dates and datetimes.

lbodor: In a library, you would try to write a generic range component, which could work with dates, times, numbers, anything with an ordering.

lbodor: I don't think we should worry about any of this though.

frankfu03: OK. I'll merge it now.

PR #776, opened 2021-01-27
Fix the error of new Date() returns invalid date in Safari

frankfu03:


frankfu03:
The creation of a new `Date` object from `new Date('2021-01-28 11:05:00')`  will get an 'invalid date' error. But Firefox, Chrome and Opera will output the correct date.
All browsers support the following formats (partial list):
```
    var d = new Date("2021/01/28 11:05:00");
    var d = new Date("2021-01-28**T**11:05:00");
```

lbodor (approved):

frankfu03:
Changes have been deployed to GNSS Site Manager DEV: https://dev.gnss-site-manager.geodesy.ga.gov.au/. You may test it using a Safari browser on any devices (eg, iPad, iPhone, etc).

